PR #2, opened 2017-03-01
refactor(radar): Converts `radar` into a computed property

pzuraq:
It turns out that `radar` needs to be a computed property in order to respond to array mutations. Something in the template itself has to be bound in order for radar to be recomputed, and as the only thing that is bound at all is `virtualComponents`, they needed to be moved into `radar` proper to put it on the refresh path.

### DIFF
--- a/addon/-private/data-view/radar/index.js
+++ b/addon/-private/data-view/radar/index.js
@@ -1,59 +1,124 @@
+import Ember from 'ember';
+
 import scheduler from 'vertical-collection/-private/scheduler';
 import Token from 'vertical-collection/-private/scheduler/token';
 
+import VirtualComponent from 'vertical-collection/-private/data-view/virtual-component';
+
 import { assert } from 'vertical-collection/-debug/helpers';
 
+const {
+  A,
+  set
+} = Ember;
+
 export default class Radar {
-  constructor({ scrollContainer, itemContainer, itemElements, totalItems, minValue, renderFromLast, parentToken }) {
-    const {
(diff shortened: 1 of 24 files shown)


addon/-private/data-view/radar/index.js
```diff
@@ -63,30 +128,156 @@ export default class Radar {
...
+    if (delta) {
+      if (delta > 0) {
+        for (let i = 0; i < delta; i++) {
+          let component = new VirtualComponent(this.token);
+          set(component, 'content', {});
+
+          virtualComponents.pushObject(component);
+        }
+      } else {
+        for (let i = virtualComponents.length; i > delta; i--) {
```
srowhani: This should be `virtualComponents.length - 1` to avoid calling `destroy` on undefined

pzuraq: Good catch, fixed that. This is the logic I was talking about for resizing by the way, it's roughly there but it doesn't really work yet.

srowhani: I noticed haha. Another thing to consider is that `div` won't fire a resize event.

Another option would be to maybe have a service run in the context of the consuming app that will listen for the window resize event, and then update any `vertical-collection`'s that have a different bounding box

pzuraq: Good point. I think we could probably use a strategy similar to the `addScrollListener` one, it's much lighter and more encapsulated than a normal service would be

PR #3, opened 2017-03-02
fix(demo): Small fixes in dummy app

srowhani:
Some small things I noticed as I was looking through the dummy application.

Hope this helps!


pzuraq:
Looks good, thanks!

PR #9, opened 2017-03-20
perf(mutation-observer): Watch for subtree modifications to schedule …

srowhani:
…remeasures

(edit) https://srowhani.github.io/ember-frost-list/

# CHANGELOG
 - perf / cache for-loops
 - on _measure, compute all orderedComponents
 - remove yielded `heightDidChange` action
 - instantiate and clean up mutation observer in `didInsertElement` and `willDestroy` life cycle hooks


runspired:
Overall this looks good. Array length caching is debateable. I have some concerns about the performance of using a mutation observer and even whether using one even enables proper resize listening (in my previous experiments it missed a lot of potential use cases), and even where it didn't miss it was often inconvenient as it's hard to separate noise from intent.

A better strategy may be:

- remeasure all components on data changes
- measure individual components on insertion
- remeasure all components on browser resize

This still misses the case where components are in a relatively sized box whose dimensions are subject to the size of a separate html subtree that may change, which was the original reason I moved to `raf` polling and later encouraged explicit actions.

Parting note: do you have any profile screenshots to share?

pzuraq:
@srowhani actually implemented the `heightDidChange` action to make mutations explicit, and we may end up having to go with that. The issue is that it isn't 100% intuitive - in his case, he has a list of items which can be expanded, and a button _outside_ of the list which expands all of items. Sending the action up from the component from a button click or checkbox change event is intuitive, but I think that having to send it on external changes isn't super intuitive, especially if those changes won't trigger lifecycle hooks (object reference with a `toggled` property, for instance).

I've been spiking on a couple of alternatives, so far I think there are three solid options:

1. We do a double measure. We can measure just before any of the scroll logic occurs, grab the existing measurements, render, and then remeasure. This is actually not terrible, you can check out the implementation on the `feature/remeasure-before-scroll` branch. It hits us about 1-1.5ms per rerender, definitely not desirable but also not noticeable (on a Macbook Pro, at least). I believe this is the most iron-clad approach in terms of correctness but it's probably also the least performant.
2. We figure out a way to trigger a `resize` event on the `itemContainer`. If a single item's height changes, the container's height will change. The only issue here would be if 2+ items changed and their cumulative height was the same (still trying to think if this could cause major issues, I don't see many edge cases where it could but I may be wrong). I have a working proof of concept I'll push in a separate PR tomorrow.
3.  We stick with explicit action sending and document it very clearly. Any/all mutations from anywhere in your app that could change _any_ state in the `vertical-collection` must be trigger the `heightDidChange` action somehow.

I'm also open to working on the `MutationObserver` approach as well, especially if option 2 doesn't pan out.

srowhani:
I found that applying a small debounce to MutObserver filters out most of the noise, as a lot of what's being picked up is the attr changes on the `vertical-collection` element to `padding-top` and `padding-bottom`.

A lot of the visual bugs I had been experiencing had been resolved in `feature/remeasure-before-scroll` There were some edge cases though, such as an item remeasure not being picked up, or scrollTop being computed incorrectly when the items being resized change the scrollTop of the parent container.

@runspired I'll link a dummy app using mutation observer with the applied debounce

runspired:
if anything for debounce it should use raf + setTimeout

@pzuraq I think we can eliminate the double pass by instead embracing separate render frames.

PR #12, opened 2017-03-23
refactor(measure): Simplifies measurement and only measures when necessary

pzuraq:
This PR simplifies the current state management for measuring VCs. They no longer cache data   about their geometry or measure state. Additionally, it changes the logic surrounding measurement in general. We now measure in three phases:

1. We measure the all items just before we make any changes or measurements. This is to capture any possible changes that could have happened while items were rendered.
2. If necessary we measure the items that could affect the scroll position post-render (e.g. if we are scroll up). This occurs after the Radar's update phase (RAF 'measure').
3. We use `run.next` to defer a full measurement of all items until after the RAF, making sure the extra work doesn't block us.

There is also a minor perf improvement, we no longer add `_prependOffset` in the sync phase. Instead, we coalesce any changes that could have occurred in the measure phase into the offset and set it once. This means that we have, at most, 1 forced layout per update, and only when we are scrolling upwards.

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -1,7 +1,10 @@
 import Radar from './index';
 import SkipList from '../skip-list';
+import Ember from 'ember';
 
 import { assert } from 'vertical-collection/-debug/helpers';
+
+const { run } = Ember;
 
 export default class DynamicRadar extends Radar {
   init(...args) {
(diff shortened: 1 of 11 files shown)


pzuraq:
This PR supercedes #9, as it guarantees that correct measurements will be made of items if their heights are updated.

pzuraq:
cc @runspired @srowhani

PR #15, opened 2017-03-24
feature(resize): Adds a resize event

pzuraq:
Even when we are always remeasuring, there is an edge case that we can't easily account for. If a large number of items are rendered, and then all of them are shrunk (or even one very large item is shrunk), it's possible that A. no scroll event is fired (sometimes it happens when we resize items, sometimes it doesn't) and B. the buffers aren't large enough to account for the delta, and we end up showing the padding below.

This adds a "resize" event based on [the concept found here](https://github.com/marcj/css-element-queries/blob/master/src/ResizeSensor.js)

The basic idea is you listen to the scroll event on a hidden container that is a child of the element you are trying to measure, with `width: 100%` and `height: 100%`. It has a child with a very large height/width, and when it resizes the scroll bars change size triggering the scroll event.

This is pretty hack-y, but it seems to be the best solution out there. It's worth noting that I had to follow the exact configuration here, with two child elements both having scroll listeners - it was possible to trip anything else up for some reason 😕 . I'm also not entirely sure how to test this, given it's based on some pretty finicky browser behavior that could change from version to version (that is, specifically what types of resizes will _not_ trigger a scroll on the `itemContainer`).

Also worth noting that this alone is not enough to fully cover measuring when elements are resized, there's the edge case where multiple items are resized but equal the same height in the end. That case may be rare enough that we could drop the pre-measurement behavior (or atleast make it optional).

### DIFF
--- a/addon/-private/data-view/radar/index.js
+++ b/addon/-private/data-view/radar/index.js
@@ -43,7 +43,6 @@ export default class Radar {
     this.bufferSize = bufferSize;
     this.renderFromLast = renderFromLast;
 
-    this._updateVirtualComponentPool();
     this.scheduleUpdate();
   }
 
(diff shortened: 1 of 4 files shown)


pzuraq:
cc @runspired

alexander-alvarez:
out of curiousity -- is this something you could leverage https://github.com/EmberSherpa/ember-element-resize-detector for?

pzuraq:
It could be, I was hoping to avoid using external libraries because most of the one's I looked at were fairly heavy (this one includes multiple resize detection strategies) and used RAF, which could interact with our usage of RAF in not-so-great ways. Another reason why we need first-class RAF support 😕

PR #25, opened 2017-04-03
docs(readme): fixed install link

luxferresum:
docs(readme): fixed install link


pzuraq:
Awesome, thanks for the fix! 😄

PR #41, opened 2017-04-16
bugfix(arrays): Fixes RecordArray behavior

pzuraq:
Turns out we can't just pull the `content` out of array proxies, as they could be doing tricky things internally and using the API to hide that (which makes sense). This PR refactors to use the standard APIs.

Additionally, this PR adds tests for RecordArrays in an acceptance test (necessary to have a functioning store for the test) and cleans up unit tests to run against all array scenarios possible at once. Fixes a couple of small bugs that were discovered because of this.


pzuraq:
Failures in the PR branch because of some weird conflict with NPM, only on the `ember-beta` scenario, going to merge and see if it goes away.

PR #42, opened 2017-04-17
bugfix(reset): Fixes the case when a collection is emptied

pzuraq:
Turns out Glimmer 1 and 2 track their end nodes slightly differently, so we had to account for that. Glimmer 2 also doesn't seem to completely delete components if you empty an `each`. Caught a couple of small bugs as well.


srowhani:
@pzuraq Just tested it out, works great! Thanks 👍

PR #44, opened 2017-04-18
refactor(padding): Use DOM elements for before/after padding

pzuraq:
This PR solves two issues, the first being that neither `padding` nor `margin` really works in all cases (inside `<table>` and `<ul>` elements, for instance), and the second being Glimmer 2's behavior when removing `each` loops.

The solution for the padding issue is to create our own custom elements, `occluded-content`, which are styled to display appropriately in whatever their container element happens to be. This appears to work perfectly for our use cases.

The `each` issue (which was discovered during implementation) is a bit more tricky. It seems that Glimmer 2 will delete all element's within an `each` if it knows that the content of the `each` is being cleared (set to an empty array). Because of our DOM manipulation, the firstNode and lastNode of the each could be anywhere and so Glimmer breaks.

The only foolproof way around this is to give Glimmer a stable beginning and end node that we never manipulate or move. This also happens to be the requirement for our `occluded-content` elements - one on either end. This PR combines the concepts and wraps the `occluded-content` elements in a VC to ensure proper boundaries across Glimmer versions.


pzuraq:
cc @runspired

PR #50, opened 2017-04-22
WIP: Guard against DOM related APIs

erkarl:
Minimum viable hack to get it working with FastBoot.

### DIFF
--- a/addon/-private/data-view/virtual-component.js
+++ b/addon/-private/data-view/virtual-component.js
@@ -5,7 +5,15 @@ import { assert } from 'vertical-collection/-debug/helpers';
 
 const { set } = Ember;
 
-const doc = document;
+let notFastBoot = () => {
+  return typeof FastBoot === 'undefined' ? true : false;
+};
+
+let doc = null;
+if (notFastBoot()) {
+  doc = document;
+}
+
 let VC_IDENTITY = 0;
 
 export default class VirtualComponent {
--- a/addon/-private/utils/element/closest.js
+++ b/addon/-private/utils/element/closest.js
@@ -1,21 +1,32 @@
 const VENDOR_MATCH_FNS = ['matches', 'webkitMatchesSelector', 'mozMatchesSelector', 'msMatchesSelector', 'oMatchesSelector'];
 let ELEMENT_MATCH_FN;
 
-VENDOR_MATCH_FNS.some((fn) => {
-  if (typeof document.body[fn] === 'function') {
-    ELEMENT_MATCH_FN = fn;
-    return true;
-  }
-  return false;
-});
-
-export default function closest(el, selector) {
-  while (el) {
-    if (el[ELEMENT_MATCH_FN](selector)) {
-      return el;
-    }
-    el = el.parentElement;
-  }
-
+let notFastBoot = () => {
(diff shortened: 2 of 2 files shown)


runspired:
Thanks for getting this started! Any chance you can help drive it to completion? Is required for us to release a 1.0 :)

erkarl:
Sure. At the moment I'm using conditionals not to use VC under FastBoot at all. I imagine a lot more things are going to break once I remove it.

https://github.com/erkarl/fronthat/blob/master/app/components/job-list/template.hbs#L1-L7

What should the default behavior be for FastBoot? I kind of like that it's rendering everything (200 in my case) because of SEO purposes. However, for bigger collections, it won't be feasible.

runspired:
@erkarl I think for fastboot we want to enable folks to set a total number to render, with the default being 1.  However, I need to discuss with @pzuraq what that would entail (I don't want a crazy refactor of the internals just to support this, needs to be pretty clean).

For now, just getting it to not error in fastboot is a huge win, even if that means rendering either all or nothing.  How does this overlap with #63 

```
{{#vertical-collection
  _fastbootRenderCount=1
}}
```

runspired:
cc @antidis

antidis:
@runspired I don't think this will solve the problem we're seeing as we are not using Fastboot.

runspired:
@antidis aaaah interesting, I had not picked up on that detail.

antidis:
@erkarl I think the `closest.js` from here will solve your problem too: https://github.com/html-next/vertical-collection/pull/63/files#diff-07bb74a75c6ee9bb9179e2fabee6f258

erkarl:
Thanks, @antidis - I'll rebase this over the weekend.

pzuraq:
Having users specify the number they want rendered in fastboot works. The only other possible choice I can think of is allowing users to manually specify `scrollContainerHeight`, which would override the actual `scrollContainerHeight` if it exists. The win there would be allowing users to change `scrollContainerHeight` dynamically like in @cibernox's case with `ember-power-select`, _without_ having to use a resize sensor, but the downside is it's more verbose and another public API we have to support.

cibernox:
Maybe what I'm about to say doesn't make sense, but could v-c, in the absence of a fixed height (that is, the container can grow and shrink with the content) , take the max-height property, as it is the maximum viewport?

pzuraq:
I think using `maxHeight` would work, if that covers your use case then we can just do that and stick with `_fastbootRenderCount` here

cibernox:
@pzuraq I believe that it will, but I won't be able to confirm until Wednesday 🌴 🏄

runspired:
I think I like a count more, because then it is dead simple to render everything or just a subset: e.g. `items.length` for everything.  This seems like something teams will want granular control over tuning, I know ours will.

pzuraq:
Nice, enjoy the waves @cibernox 😄 

Ok, let's go with the granular API for fastboot. I'll add the `maxHeight` property to determining scrollContainer height separately and test it out in `ember-power-select-collection`

pzuraq:
This has been superceded by #68 and #63 in terms of getting it to not break when Fastboot is installed. Going to close for now, but if you still want to help with getting Fastboot working better @erkarl, the PR would definitely be welcome!

PR #60, opened 2017-04-28
Fix #57

### LINKED ISSUE #57: Can't scroll to bottom on dbMon example in Chrome
When scrolled down to the bottom-most , scroll stuck , can't get go the bottom.
screenshot_20170426_181913

runspired:
Fixes #57

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -92,7 +92,6 @@ export default class Radar {
 
       this.schedule('measure', () => {
         if (this._prependOffset !== 0) {
-          console.log('setting offset for prepend');
           this.scrollTop += this._prependOffset;
           this._prependOffset = 0;
         }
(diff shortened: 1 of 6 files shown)


runspired:
Tests are red but between both runs all versions were green, we really need to figure out this flakiness :'(

addon/-private/data-view/radar/radar.js
```diff
@@ -289,6 +289,21 @@ export default class Radar {
     }
   }
 
+  _updatePoolContents() {
```
pzuraq: We can remove this function. `_updateVirtualComponents` will always update all of the items in the pool, we just need to make sure that `itemDelta` is zero if we don't want it to jump around (like it was in the dbmon example).

runspired: see later fix :P

addon/-private/data-view/radar/radar.js
```diff
@@ -307,12 +322,16 @@ export default class Radar {
     this.scheduleUpdate();
   }
 
-  resetItems(items) {
+  updateItems(items, isReset = false) {
```
pzuraq: Renaming this function has implications for DynamicRadar, it overrides the `resetItems` function and calls super. We either need to roll that logic into here somehow, or we need to rethink the reset logic.

runspired: I fixed it there too

addon/-private/data-view/radar/radar.js
pzuraq: 👍

PR #63, opened 2017-04-28
Make closest not require document.body

antidis:
Allows us to keep our JS in `<HEAD>`.

### DIFF
--- a/addon/-private/utils/element/closest.js
+++ b/addon/-private/utils/element/closest.js
@@ -1,15 +1,18 @@
 const VENDOR_MATCH_FNS = ['matches', 'webkitMatchesSelector', 'mozMatchesSelector', 'msMatchesSelector', 'oMatchesSelector'];
 let ELEMENT_MATCH_FN;
 
-VENDOR_MATCH_FNS.some((fn) => {
-  if (typeof document.body[fn] === 'function') {
-    ELEMENT_MATCH_FN = fn;
-    return true;
-  }
-  return false;
-});
+function setElementMatchFn(el) {
+  VENDOR_MATCH_FNS.forEach((fn) => {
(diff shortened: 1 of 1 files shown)


addon/-private/utils/element/closest.js
```diff
@@ -1,15 +1,18 @@
...
-VENDOR_MATCH_FNS.some((fn) => {
-  if (typeof document.body[fn] === 'function') {
-    ELEMENT_MATCH_FN = fn;
-    return true;
-  }
-  return false;
-});
+function setElementMatchFn(el) {
+  VENDOR_MATCH_FNS.forEach((fn) => {
+    if ((ELEMENT_MATCH_FN === undefined) && (typeof el[fn] === 'function')) {
```
antidis: Prefers the first match.

addon/-private/utils/element/closest.js
```diff
@@ -1,15 +1,18 @@
...
+      ELEMENT_MATCH_FN = fn;
+    }
+  });
+}
 
 export default function closest(el, selector) {
+  if (ELEMENT_MATCH_FN === undefined) {
+    setElementMatchFn(el);
+  }
```
antidis: By basing it off `el`, and by placing it here, we don't run our check until we actually use the function the first time

There is a slight slowdown here when the function is called, but it should be marginal on the second and subsequent calls.

antidis: And this function is only used once in the component anyway, so there should be effectively no slowdown.

antidis:
cc @runspired @gavinjoyce

antidis:
Tests are failing in CI, but the same tests fail for me in `master` locally right now.

antidis:
Looks like master broke here: https://travis-ci.org/html-next/vertical-collection/builds/226719240

runspired:
Accidentally closed this via the commit.

PR #65, opened 2017-04-29
Svelte/rollup private

runspired:
~ 1KB gzipped 20KB not gzipped of savings (we are now 6.4kb total gzipped)

Looking through where our cost centers are to file-weight, I suspect there are a few things we could do to organize the concepts in `-private` better which would yield ~2.5KB in savings plus a further reduction in memory pressure from closures and scopes.


runspired:
@pzuraq I think our build errors are because we didn't configure ember-try correctly

pzuraq:
I wonder if they're related to the custom stuff installed for Chrome, since I set up the tests before headless Chrome was a thing. We should switch to headless Chrome either way.

Also Travis isn't using Yarn, we should have it do that. These failures do look legit though, NPM having issues with peerDeps?

pzuraq:
other than the test failures it looks good though 👍

PR #67, opened 2017-04-29
make resize demo fail if needed

runspired:
Still need to figure out what expected behavior is for enabling % heights etc.


pzuraq:
What's the purpose of this PR?

runspired:
It  sets up the CSS to cause the Min height error to trigger on many resizes,  which we can use to test different scenarios

Issue #70, opened 2017-04-29
Expose the key used for an item when sending an action


runspired:
See https://github.com/runspired/smoke-and-mirrors/issues/154

pzuraq:
Should we just pass this in as a third parameter to the actions?

PR #75, opened 2017-05-04
refactor(actions): Makes sure first/lastReached fire only once

### LINKED ISSUE #52: `lastReached` and `firstReached` can continue firing forever
We don't currently guard against `lastReached` and `firstReached` firing more than once for the same item, so they can keep firing after appends/prepends that don't actually affect what is rendered.

pzuraq:
Fixes #52 and #70 

* Adds `firstReached` and `lastReached` key maps that track whether or not we've sent the first/last reached actions for an item. We need separate maps for the case where we have 1 item.
* Moves the handling sending items into the Radar class itself. This is a more natural fit place for the state to live, and it also opens up the door to further refactors in the `vertical-collection` component itself.
* Refactors action sending to schedule independently for each action. This means that if a user only has the `lastReached` action, we won't be using `setTimeout` every time the first/last visible items change, for instance.
* Adds extra properties in constructors to avoid thrashing prototypes

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -1,14 +1,27 @@
-import Radar from './radar';
+import { default as Radar, NULL_INDEX } from './radar';
 import SkipList from '../skip-list';
 
 import { assert } from 'vertical-collection/-debug/helpers';
 
 export default class DynamicRadar extends Radar {
+  constructor() {
+    super();
+
+    this._firstItemIndex = NULL_INDEX;
+    this._lastItemIndex = NULL_INDEX;
+
+    this._totalBefore = 0;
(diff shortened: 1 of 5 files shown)


runspired:
@pzuraq I'd already fixed this I'm pretty sure

pzuraq:
Yeah, it turns out the bug was expected behavior (if an append ends up rendering all items anyways, it'll keep firing, which was what happened). 

This actually formalizes that first/lastReached will only fire once per item though, by tracking the keys. Before if you scrolled back and scrolled forward again the event would fire again.

runspired:
@pzuraq before the keyForItem was broken but I fixed that in #60, I think we just needed to switch from looking at index to looking at key

pzuraq:
Right, that's what this PR does

PR #77, opened 2017-05-05
Feat/make buffer size static

pzuraq:
Makes `bufferSize` a static number, with a default of 5 (10 extra components in general). Open to bikeshedding on the default value, it may make more sense to set it to 0 and let users decide what makes sense for them.

Also updates the tests, and makes sure that `StaticRadar` has the same result as `DynamicRadar` for given offset (if heights/settings are identical).

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -348,10 +348,10 @@ export default class Radar {
     } = this;
 
     // The total number of components is determined by the minimum number required to span the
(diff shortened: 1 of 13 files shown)


pzuraq:
Decided to go with a default bufferSize of 0. It makes more sense for dynamic collections because the average height will likely be above the min, so there's already a buffer, and in general I don't think we can assume what a reasonable buffer is (10 components would be overkill if only 2 are rendered at a time, for instance).

PR #79, opened 2017-05-05
bugfix(skiplist): Use Float32Arrays in the SkipList

### LINKED ISSUE #76: SkipList breaks when heights are not integers
Because the SkipList uses integer typed arrays as it's underlying store, it breaks if items are not pure integers. See the Scrollable Body example and scroll to the bottom to see this in action - it keeps detecting a difference between the actual height and the stored height, and it adds the delta to the total which then becomes inaccurate.

We could round out the values before we add them, but I think there could be small glitches due to the differences between actual heights and measured heights. Using Float32Arrays instead of UInt32Arrays is safer, and would only increase the memory usage for the bottom layer (which is currently a UInt16Array).

pzuraq:
Uses Float32Arrays to get sub-pixel accuracy from measurements. Fixes #76

### DIFF
--- a/addon/-private/data-view/skip-list.js
+++ b/addon/-private/data-view/skip-list.js
@@ -15,7 +15,7 @@ import { assert } from 'vertical-collection/-debug/helpers';
  *
  * This data structure acts somewhat like a Binary Search Tree. Given a list of size n, the
  * retreival time for the index is O(log n) and the update time should any values change is
- * O(log n). The space complexity is O(n log n) in bytes (using Uint16/32Arrays helps a lot
+ * O(log n). The space complexity is O(n log n) in bytes (using Float32Arrays helps a lot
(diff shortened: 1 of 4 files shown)


addon/-private/data-view/skip-list.js
```diff
@@ -26,7 +26,7 @@ import { assert } from 'vertical-collection/-debug/helpers';
 
 export default class SkipList {
   constructor(length, defaultValue) {
-    const values = new Uint16Array(new ArrayBuffer(length * 2));
+    const values = new Float32Array(new ArrayBuffer(length * 4));
```
runspired: I think we need to detect support for this and fallback

runspired: nvm I don't care about IE9 :P

PR #80, opened 2017-05-07
yay! :)

runspired:


runspired:
Test failures are flakey and differ on runs, although the 1.11 failure may also speak to us needing to adjust our caching to ensure a slightly different ember version.

pzuraq:
Probably an issue with an addon that has an instance initializer, 1.11 didn't have them yet. You have to remove the addon in the ember-try config for that addon. Could be ember-fastboot, since it was moved from dependencies to devDependencies.

PR #82, opened 2017-05-07
fix initial scroll offset and add test

runspired:


runspired:
Failure was the same 1.11 instance initializer that needs fixed.

PR #83, opened 2017-05-07
bugfix(ember-try): Makes ember-cli-fastboot dependency null

pzuraq:
Should fix tests

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -18,12 +18,10 @@ module.exports = {
       npm: {
         devDependencies: {
           'ember-cli-shims': null,
+          'ember-cli-fastboot': null,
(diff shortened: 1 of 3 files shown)


pzuraq:
@runspired pre-1.13.0 Ember Data didn't use JSON API internally either, so this won't work (ran the tests locally to verify since Travis is bugging out). I'm ok with dropping the beta version of Ember Data from the tests, we can sub in 1.13.0 (it should work with Ember 1.11), or we can maybe format the payload differently depending on the version of Ember Data.

runspired:
@pzuraq seriously? I think we should just set ember-data to 1.13 in this test then.

PR #84, opened 2017-05-08
Fix dynamic start with offset

runspired:
This adds a (was failing) test for starting a dynamically sized list in a position offset from the top of the scroll-container.

However, this causes the skip-list to need to deal with negative numbers. To handle this, I then changed the determination of `middleVisibleValue` to:

```
let top = this.visibleTop;

if (top < 0) {
  top = 0;
}

const middleVisibleValue = top + (this.scrollContainerHeight / 2);
```

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -83,7 +83,7 @@ export default class DynamicRadar extends Radar {
     const itemDelta = (prevFirstItemIndex !== null) ? firstItemIndex - prevFirstItemIndex : 0;
     const numCulled = Math.abs(itemDelta % numComponents);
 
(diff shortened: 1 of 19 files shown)


pzuraq:
Yeah, we should keep `visibleTop` here. `scrollTopOffset` is still needed, and `prependOffset` is whatever the offset we get when we do a prepend (in a prepend, we don't actually set scrollTop until _after_ we've rendered to avoid setting it more than once, so we need to make our calculations with that in mind). 

The correct behavior here is if `visibleTop <= 0`, we would render the first n elements (where n is the number of VCs). `visibleTop` is supposed to give us the ability to have a "negative scrollTop", which is useful for calculations and tracking state but makes no sense here. So it should probably be:

```
const middleVisibleValue = Math.max(this.visibleTop, 0) + (this.scrollContainerHeight / 2);
```

PR #89, opened 2017-05-14
[WIP] refactor(radar): startup and resize

pzuraq:
This PR does a few things mostly centered around normalizing the `scheduleUpdate` function so that we are:

1. Deferring all work into a RAF, both so measurements are much lower cost and so we don't block the critical path to rendering
2. Always remeasuring so that the internal values we run against are always up to date (`scrollContainerHeight`, `scrollTopOffset`, etc).
3. Always updating everything, including the VC pool

It drops the confusing system of getters/setters for these values and replaces them with a pipeline where every stage of rendering is clearly defined and always happens at a specific point, giving us direct control over where and when we decide to measure and allowing us to prevent forced layouts much more easily.

Still working on this, I'd like to figure out a way to avoid the extra `run` that's been added in `updateVirtualComponentPool`. The issue there is that we want to update the number of VCs if needed, but if we do we can't move _any_ of the VCs around until the new ones render, otherwise it'll cause issues with Glimmer. We can eat the cost of the second `run`, or rerender every component if the VC pool updates, or possibly try to stop doing the DOM manipulation at all.

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -222,20 +222,20 @@ export default class DynamicRadar extends Radar {
     }
   }
 
-  prepend(items, numPrepended) {
-    super.prepend(items, numPrepended);
+  prepend(numPrepended) {
+    super.prepend(numPrepended);
 
     this.skipList.prepend(numPrepended);
   }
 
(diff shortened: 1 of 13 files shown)


runspired:
Will take a look at this tomorrow :)

PR #90, opened 2017-05-17
bugfix(skiplist): Don't use TypedArray.fill unless it exists

pzuraq:


addon/-private/data-view/skip-list.js
```diff
@@ -24,15 +24,18 @@ import { assert, stripInProduction } from 'vertical-collection/-debug/helpers';
  * traverse to get the total value before and after the final index.
  */
 
+function fill(array, ...args) {
+  if (typeof array.fill === 'function') {
+    array.fill(...args);
+  } else {
+    Array.prototype.fill.call(array, args);
+  }
+}
```
runspired: I'd peek at how we did this in Heimdall

PR #94, opened 2017-05-23
perf(initial render): Adds scroll handlers inside the initial RAF

pzuraq:
The ScrollHandler has to read the initial state of the scrollContainer (scrollTop + scrollLeft) to be able to know if a scroll event has actually caused a change. Doing that directly in `didInsertElement` will cause a forced layout, so it's a better idea to schedule it inside of the first RAF.

Timeline Before:

screen shot 2017-05-23 at 12 09 50 pm

Timelines After: 

screen shot 2017-05-23 at 12 08 35 pm

screen shot 2017-05-23 at 12 09 06 pm

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -168,9 +168,9 @@ const VerticalCollection = Component.extend({
     // Initialize the Radar and set the scroll state
     this._initializeRadar();
     this._initializeScrollState();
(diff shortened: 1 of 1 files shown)


erkarl:
Looks good. These screenshots are amazing with ember-perf-timeline enabled :)

runspired:
I'm overall 👍 on this change

PR #99, opened 2017-06-03
feature(init): Adds initialRenderCount option

pzuraq:
Adds an `initialRenderCount` option which will render a static number of items on the first pass. This allows users to: 

A. Not take the initial perf hit on first render from rendering so many items inside of a RAF and
B. Specify a number to be rendered for things like Fastboot

Timelines before:

screen shot 2017-06-02 at 4 57 05 pm

screen shot 2017-06-02 at 4 57 02 pm

Timelines after:

screen shot 2017-06-02 at 4 57 42 pm

screen shot 2017-06-02 at 4 57 34 pm

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -4,8 +4,8 @@ import SkipList from '../skip-list';
 import { assert, stripInProduction } from 'vertical-collection/-debug/helpers';
 
 export default class DynamicRadar extends Radar {
-  constructor() {
-    super();
+  constructor(parentToken, initialItems, initialRenderCount, startingIndex) {
+    super(parentToken, initialItems, initialRenderCount, startingIndex);
 
     this._firstItemIndex = NULL_INDEX;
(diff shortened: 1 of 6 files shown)


addon/-private/data-view/radar/dynamic-radar.js
```diff
@@ -4,8 +4,8 @@ import SkipList from '../skip-list';
 import { assert, stripInProduction } from 'vertical-collection/-debug/helpers';
 
 export default class DynamicRadar extends Radar {
-  constructor() {
-    super();
+  constructor(...args) {
```
runspired: use explicit argument names instead of spread

addon/-private/data-view/radar/static-radar.js
```diff
@@ -2,8 +2,8 @@ import { default as Radar, NULL_INDEX } from './radar';
 import { stripInProduction } from 'vertical-collection/-debug/helpers';
 
 export default class StaticRadar extends Radar {
-  constructor() {
-    super();
+  constructor(...args) {
```
runspired: use explicit argument names

Issue #74, opened 2017-05-02
Vertical collection does not work very well with IE11 and Edge


jevanlingen:
The plugin is just great in Chrome and Firefox. Sadly if I watch the [table demo](https://html-next.github.io/vertical-collection/#/examples/dbmon) in the Windows browsers, not so much.

**Internet Explorer 11**
First entries are loaded, but when scrolling down no new entries are drawn:
![image](https://cloud.githubusercontent.com/assets/1281763/25610101/2fdd14d6-2f22-11e7-9bd6-c1a87a398588.png)

**Edge**
Only first entry row is drawn (when scrolling up and down, all entries are drawn, so it seems some initialization error):

![image](https://cloud.githubusercontent.com/assets/1281763/25610156/6dd2b6a6-2f22-11e7-809f-7fa5f8ccbea5.png)

runspired:
Does this only happen on the table demo? Or on the other demos as well?

jevanlingen:
**Internet Explorer**
- Infinite Scroll: Same behaviour
- Flexible Layout: Demo does not start, only gray square is visible:
![image](https://cloud.githubusercontent.com/assets/1281763/25650983/c5b9e2e0-2fe0-11e7-8e70-842ea4673a09.png)
- Scrollable Body: Demo does not start, whole page stays white:
![image](https://cloud.githubusercontent.com/assets/1281763/25650996/da5eed44-2fe0-11e7-9054-b47eacc36c39.png)

**Edge**
- Other demos do work fine

pzuraq:
The issues with Edge are definitely related to the way tables render. Specifically, something to do with the way that we're adding VCs on initial render. If we populate the VC pool when we create the Radar, it renders like it should.

I can't seem to exactly reproduce the IE11 issues. For me, it renders the items, but doesn't update them on scroll.

Gotta handle some of the other issues, but edge and IE support are definitely a priority so will come back to this asap.

pzuraq:
IE11 should have been fixed by #92, still working on Edge so keeping this issue open

kumkanillam:
In IE-11 , I get the error `Unable to get property 'scrollTop' of undefined or null reference`.

![image](https://cloud.githubusercontent.com/assets/16604113/26481963/0caac042-4202-11e7-87fa-c2bf7bf289b1.png)

pzuraq:
It looks like you're using an older version, can you update to the latest master and see if it still happens? We should be releasing a new beta soon

[...]

PR #104, opened 2017-07-15
bugfix(M$): Fixes measurement in Microsoft browsers

pzuraq:
Swaps out the range based measurement strategy for one that measures elements. This means that all content _must_ be wrapped in top level elements. Also includes some test cleanup so they'll run in IE (we need to get cross browser automated tests going).


pzuraq:
Fixes #74

PR #107, opened 2017-07-17
feat(recycling): Makes recycling optional

pzuraq:
Adds the `shouldRecycle` option which, when false, turns off component recycling. This may make sense for certain use cases, but it should be noted that in early testing this strategy is a major perf hit.

With recycling:

screen shot 2017-07-17 at 9 12 22 am

Without recycling:

screen shot 2017-07-17 at 9 12 04 am

screen shot 2017-07-17 at 9 12 08 am

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -22,13 +22,14 @@ const {
 export const NULL_INDEX = -2;
 
 export default class Radar {
-  constructor(parentToken, initialItems, initialRenderCount, startingIndex) {
+  constructor(parentToken, initialItems, initialRenderCount, startingIndex, shouldRecycle) {
     this.token = new Token(parentToken);
 
     this.items = initialItems;
     this.estimateHeight = 0;
     this.bufferSize = 0;
     this.startingIndex = startingIndex;
+    this.shouldRecycle = shouldRecycle;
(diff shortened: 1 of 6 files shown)


addon/-private/data-view/radar/radar.js
```diff
@@ -259,19 +264,35 @@ export default class Radar {
...
+    if (_wasReset) {
+      for (let i = 0; i < orderedComponents.length; i++) {
+        if (shouldRecycle === true) {
+          // If the underlying array has changed, the indexes could be the same but
+          // the content may have changed, so recycle the remaining components
+          const component = orderedComponents[i];
+          component.recycle(objectAt(items, component.index), component.index);
+        } else {
+          // If recycling is disabled we need to delete all components and clear the array
+          _componentPool.push(orderedComponents.shift());
```
srowhani: Will this successfully clear all components? Not sure about the effects of resizing an array while iterating over it.

```
a= [1,2,3,4,5]
> (5) [1, 2, 3, 4, 5]
for (let i = 0; i < a.length ; i++) {
  console.log(a.shift());
}
> 1
> 2
> 3
> undefined
```

pzuraq: Good catch! Yeah, will need to update that. Should have marked this branch as WIP too, still needs tests.

PR #119, opened 2017-08-09
bugfix(mutation): Handle scenario where items change position

g-cassie:
If you had a list of items `[1, 2, 3, 4, 5]` and it was changed to `[1,
4, 3, 2, 5]`, `vertical-collection` would continue to render the items
in their original position.  

This fixes the bug as discussed with @pzuraq on the Ember Slack.


pzuraq:
Looks great! Will merge once tests pass, thanks again for taking this on!

PR #123, opened 2017-08-16
Check null for _debugDidUpdate

billy-addepar:
@pzuraq


pzuraq:
Awesome, looks good. Will merge after tests pass

PR #125, opened 2017-08-17
Add 0 checking for visible top & container height

billy-addepar:
I have an issue in VC usage that the container height is 0 at the beginning (because of height 100%) but will be updated later on. When the container height is 0, both `visibleTop` and `_scrollContainerHeight` are 0 and hence `visibleBottom` is -1. This crashes the app because the skip list expects item index to be non-negative. 

The fix is to set `visibleBottom` to be 0 when both `visibleTop` and `_scrollContainerHeight` are 0.

@pzuraq @runspired

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -534,14 +534,10 @@ export default class Radar {
   }
 
   get visibleBottom() {
-    // This is the case where the container of this vertical collection could have height 0px at
+    // There is a case where the container of this vertical collection could have height 0px at
     // initial render step but will be updated later. We want to return visibleBottom to be 0 rather
     // than -1.
-    if (this.visibleTop == 0 && this._scrollContainerHeight == 0) {
-      return 0;
-    }
-
-    return this.visibleTop + this._scrollContainerHeight - 1;
+    return Math.max(this.visibleTop + this._scrollContainerHeight - 1, 0);
   }
 
   get totalItems() {
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -534,6 +534,13 @@ export default class Radar {
   }
 
   get visibleBottom() {
+    // This is the case where the container of this vertical collection could have height 0 at
+    // initial render step but will be updated later. We want to return visibleBottom to be 0 rather
+    // than -1.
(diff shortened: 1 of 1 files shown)


addon/-private/data-view/radar/radar.js
```diff
@@ -534,6 +534,13 @@ export default class Radar {
...
 
   get visibleBottom() {
+    // This is the case where the container of this vertical collection could have height 0px at
+    // initial render step but will be updated later. We want to return visibleBottom to be 0 rather
+    // than -1.
+    if (this.visibleTop == 0 && this._scrollContainerHeight == 0) {
+      return 0;
+    }
+
     return this.visibleTop + this._scrollContainerHeight - 1;
```
pzuraq: I think this can be simplified as `return Math.max(this.visibleTop + this._scrollContainerHeight - 1, 0);`

addon/-private/data-view/radar/radar.js
```diff
@@ -534,6 +534,13 @@ export default class Radar {
   }
 
   get visibleBottom() {
+    // This is the case where the container of this vertical collection could have height 0px at
+    // initial render step but will be updated later. We want to return visibleBottom to be 0 rather
+    // than -1.
+    if (this.visibleTop == 0 && this._scrollContainerHeight == 0) {
```
pzuraq: always use triple equals

runspired:
> I have an issue in VC usage that the container height is 0 at the beginning (because of height 100%) but will be updated later on. When the container height is 0, both visibleTop and _scrollContainerHeight are 0 and hence visibleBottom is -1. This crashes the app because the skip list expects item index to be non-negative.

If I understand correctly, is this to address situations in which the container size is initially unknowable but has some form of maxHeight?

pzuraq:
Yes, specifically users may use JS to set the height of the container (or its parent, and the container has `height: 100%`, etc).

runspired:
I feel 2/10 that there are better solutions outside of this lib but this is a small enough change in okay with it. May want additional code comment docs and/or tests to ensure the "why" is preserved.

pzuraq:
Agree, it's mainly that this is a behavioral error - it doesn't make sense to have a `visibleBottom` that is less than 0. If we want to enforce users have a scroll container with some height, we should add an assertion to the measurement phase in Radar.

PR #126, opened 2017-08-23
[Perf] (WIP): Replace Skip List by Red Black Tree

billy-addepar:
# TL; DR
This PR significantly improves performance of vertical collection by replacing skip list with [red black tree](https://en.wikipedia.org/wiki/Red%E2%80%93black_tree) (RBT). It's faster because: 
- Initialization time goes down from O(N) in skip list to O(1) for RB tree.
- Memory usage is O(N) in worse case (same with skip list) but majoriy of cases, the tree uses sub-linear mem usage. In many cases, it's just O(1) !!
- Prepend & append new elements in the tree take O(logN) instead of O(N) for skip list, making infinite scrolling faster.
- Finding partial sum takes O(logN), same with skip list.
- The RB tree implementation is fully functional with demo app. There might be some code I need to clean or tests to add but overall it's good to play around with.

## The partial sum problem
One problem arises from vertical collection implementation is to find a good data structure to efficiently update & retrieve partial sum of an array. The problem could be sumarized as follow:

Let A[1..n] be an array of real numbers. Design an algorithm to perform any sequence of the following operations:
- update(index, value) -- Update position index with a value. 
- find(targetSum): returns the maximum number `i` such that sum(A[0]..A[i]) <= targetSum < sum(A[0]..A[i + 1]). 

There are multiple ways to solve this problem in log(N) time. Common approach is to use some binary search tree. Indeed, the skip list in the current implementation is actually a [binary segment tree](https://en.wikipedia.org/wiki/Segment_tree). Using skip list has a number of advantages, including simple implementation, retrieval & updating take logarith time. However, the biggest issue with skip list is the O(N) initialize time and it's a static array. This means that to support deletion & insertion we have to tear down entire array and recreate it again. 

## Self Balanced Binary Tree Approach. 
The partial sum problem can be solved using any self balanced binary search tree. The algorithm is as follow:
- The search key for the tree is the index in the array. Nodes with indexes bigger than root's index is on the right of the root and vice versa. In addition to that, we store the `sum of left subtree` of any node in the node data, including the node itself. It's important to note that we only store sum of the left subtree because nodes in the right subtree have bigger index.
- update(index, value): we traverse from the root to the node that has index. Update its left tree sum and all of its parents whose left subtree also contains that node.
- find(targetSum): We traverse from the root. If the `targetSum` is bigger that left subtree sum, go to the right. If it's smaller or equal than `strict left subtree` sum, traverse to the left.
- Prepend & Append items take O(logN) as a normal self balanced binary search tree.

Main advatages of using BST is that it's a dynamic structure. This means that you can delay tree initialization until we only need to process actual data. Deletion & insertion items take only O(logN). More importantly, it allows use to have sub-linear memory usage on average like described in the section below.

## Sub-linear memory usage. 
How can we have sub-linear usage if we have at least N items?
The answer is at the nature of vertical collection (VC). Main usage of VC is for list or table data. For long list with more than 10k items, most items or rows have the same height. Instead of storing single item index in the node, we store a continuous block of items with same height. Each node in the tree represents an interval `[start..end]` for items with same height. It keeps `start`, `end` and `height` variable. The tree still complies with its BST restriction but we have much less nodes in the tree and hence faster searching & updating. 

Updating an index might require splitting a node into 3 different nodes. For example, if the interval is [6 6 6 6 6] and we want to update middle index with 5, the node is splited into 3 nodes [6 6] [5] [6 6]. This operation is the same like inserting new node into a tree and takes O(logN)

As VC use estimate height for array initialization, we can assume that all items have same height at the beginning. If all rows in the list/ table have the same height, we have only 1 node or O(1) mem usage!!! Indeed, I believe that any list with more than 10k items will very likely have item with same height. 

The last question we have to answer is what self balance BST tree should we use? 

## Red Black tree vs AVL tree vs other self balance BST
I have considered a number of trees for implementation. My initial approach was to use [AVL tree](https://en.wikipedia.org/wiki/AVL_tree) because it has faster retrieval time. As I observe, we indeed have more node insertion & deletion than query. This is especially true when we use an interval `[start..end]` in a node and updating an index falling in between the interval with different height means splitting up the interval into 3 sub intervals. AVL is more rigidly balanced and hence provide faster look-ups. RB has smaller number of tree rotation on average and hence provide a better performance. In fact, most common std library uses RB tree for their map & set implementation.

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -1,7 +1,7 @@
 import Radar from './radar';
-import SkipList from '../skip-list';
-// import SkipList from '../rb-tree/rb-tree-wrapper';
-import RbTreeWrapper from '../rb-tree/rb-tree-wrapper';
+// import SkipList from '../skip-list';
+import SkipList from '../rb-tree/rb-tree-wrapper';
+// import RbTreeWrapper from '../rb-tree/rb-tree-wrapper';
 import roundTo from '../utils/round-to';
 
 import { stripInProduction } from 'vertical-collection/-debug/helpers';
@@ -36,10 +36,10 @@ export default class DynamicRadar extends Radar {
     // Create the SkipList only after the estimateHeight has been calculated the first time
     if (this.skipList === null) {
       this.skipList = new SkipList(this.totalItems, this._estimateHeight);
-      this.rbTreeWrapper = new RbTreeWrapper(this.totalItems, this._estimateHeight);
+      // this.rbTreeWrapper = new RbTreeWrapper(this.totalItems, this._estimateHeight);
     } else {
       this.skipList.defaultValue = this._estimateHeight;
-      this.rbTreeWrapper.defaultValue = this._estimateHeight;
+      // this.rbTreeWrapper.defaultValue = this._estimateHeight;
     }
   }
 
@@ -74,16 +74,16 @@ export default class DynamicRadar extends Radar {
     // const { values } = skipList;
 
     let { totalBefore, totalAfter, index: middleItemIndex } = this.skipList.find(visibleMiddle);
-    let { totalBefore: before2, totalAfter: after2, index: index2 } = this.rbTreeWrapper.find(visibleMiddle);
-    if (index2 === middleItemIndex) {
-      console.log('Equal ' + index2);
-    }
-    if (before2 !== totalBefore || after2 !== totalAfter || index2 !== middleItemIndex) {
-      console.log(middleItemIndex, index2, totalBefore, before2);
-      // debugger
-      // this.rbTreeWrapper.find(visibleMiddle);
-      // throw new Error('Not matching')
-    }
+    // let { totalBefore: before2, totalAfter: after2, index: index2 } = this.rbTreeWrapper.find(visibleMiddle);
+    // if (index2 === middleItemIndex) {
+    //   console.log('Equal ' + index2);
+    // }
+    // if (before2 !== totalBefore || after2 !== totalAfter || index2 !== middleItemIndex) {
+    //   console.log(middleItemIndex, index2, totalBefore, before2);
+    //   // debugger
+    //   // this.rbTreeWrapper.find(visibleMiddle);
+    //   // throw new Error('Not matching')
+    // }
 
     const maxIndex = totalItems - 1;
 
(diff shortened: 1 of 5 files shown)


billy-addepar:
@runspired @pzuraq

pzuraq:
Any updates to low level constructs in this library for performance reasons must show clear, objective results in the form of low level perf benchmarks. We haven't yet setup the proper performance test suite for this level of optimization because it's not a priority, so you'll have to do that yourself if you'd like to get this merged.

If perf differences are clear and noticeable in Chrome timelines or memory snapshots that may be enough. Otherwise, check out https://github.com/krisselden/ember-macro-benchmark to see some of the low level types of testing we would expect to see to prove that this is better.

runspired:
At first glance I think this may be a great direction to explore. My primary fear with this though is that we cannot always key by index nor do we always have knowledge of partial updates to the data given to us.  The collection manager proposal and allowing custom iterators would enable us to give the collection that knowledge.

pzuraq:
Reading through the description I like the direction this is heading. If we can get objective benchmarks that prove this is better in general then I'm all for moving forward with it.

We may even be able to optimize further by using [WebAssembly](http://webassembly.org/) to manually manage memory and the low level operations here. This will likely provide the best perf of all, since we'll be able to use the superior data structure with the benefits of TypedArray-like memory management and speed.

But in the end, we won't know until we benchmark it, that's the key here.

pzuraq:
@runspired can you link to the collection manager proposal? Not sure which one you're referencing

billy-addepar:
Sorry for the rush for this WIP. I should not have a WIP PR in the open source world. I took a day break to work on this before getting back to my normal work. That's why there are some debugging code here. The RB code part is pretty solid though since I have run tests with it. 

I will do the benchmark as suggested.

pzuraq:
WIP isn't an issue, it's actually good for transparency and for getting ideas out there 🙂 

The main thing is that we aren't quite ready to focus on low-level performance, we haven't yet shipped `v1.0.0` and we need to make sure that everything is working as is and fully tested, so @runspired and I don't really have time to build out the perf tests we would need to verify that this is a better strategy. If you can also add those tests then this PR is great, although I do also want to follow through on the concern about collection managers/keying by index.

runspired:
@billy-addepar WIP is always welcome if labeled as such and even though this wasn't your description did make it clear where you knew it to be incomplete.  Overall this is an awesome PR, but due to the technical and fundamental nature of the changes it's likely to take a while and iteration to get fully reviewed and approved.  I'm excited to work with you on it!

billy-addepar:
I think this can go in after 1.0 version. We should focus on stability as 1.0 release is coming. I will be kept busy with other and this would take a few weeks to go in. 

I am very curious the collection manager proposal. If you can link it here, it would be great. Otherwise, we can talk offline about it.

billy-addepar:
I tested current skipList with 200k items and it seems not have any noticeable perf difference with RB tree. With 500k, there might be some tiny flickering when adding item but it is still usable. With 1M the list cannot render as the total sum exceeds maximum float number limit

Closing this issue for now.

Issue #127, opened 2017-08-26
VC does not detect table body height change.


billy-addepar:
My table is using VC and only 2 rows are rendered at the beginning:

<img width="1043" alt="screen shot 2017-08-26 at 2 07 22 pm" src="https://user-images.githubusercontent.com/11281370/29745111-e94f3660-8a67-11e7-8c57-b9e7ed8884c9.png">

You can see entire white space below the first 2 rows. However, as you start scrolling all other rows start to render.
<img width="1036" alt="screen shot 2017-08-26 at 2 08 28 pm" src="https://user-images.githubusercontent.com/11281370/29745118-0e5cedd0-8a68-11e7-8c7a-1ee8880f1b89.png">

To reproduce this:
- Checkout my branch https://github.com/billy-addepar/vertical-collection/tree/billy/table-reproduction
- Go to route http://localhost:4200/#/examples/table

This happens because in the first RAF execution, the table only measures the header and the body height is 0. VC detects 0 container height and only renders 2 items. However, as the body has something inside, the table body expands to its full height leaving entire blank space below 2 rows. This is not a resize event. VC does not detect change in body height and does not update accordingly. 

When you start scrolling, the VC sees the new container height and renders all rows as expected. All these events could be observed by putting a debugger in the radar. 

Note that this only happens for my table since I have some special CSS applied to table. (check app.css). 

Not sure how to fix this as I cannot find an easy way to detect change in table body height.

pzuraq:
I think this may be more of a CSS/timing issue, the container should be able to know it's height by the first render frame somehow. As a temporary fix, consider using initialRenderCount to fill out the first render.

billy-addepar:
That sounds a good idea. The only issue with that is `initialRenderCount` is used very early in the construction of VC component
https://github.com/html-next/vertical-collection/blob/master/addon/-private/data-view/radar/radar.js#L67

At that time, we do not have everything rendered on screen yet. Nor do we know the height of table. Hence it's very hard to know what `initialRenderCount` value should be

billy-addepar:
I should we should consider moving some variables initialization in `init` of `radar` to the first RAF. At that time the parent component of the VC has some knowledge of surrounding environment and can give more accurate estimate of `bufferSize`, `initialRenderCount`, `estimateHeight`, etc.

pzuraq:
[...]

PR #128, opened 2017-08-28
(bugfix): set correct max height.

billy-addepar:
In javascript `parseInt (100%)` does not return NaN. It returns number 100 instead (yay javascript!).

This PR fixes the logic to get max height in radar with percentage. Since `max-height` percentage represents the maximum amount of space that an element can have relative to its parent, we should extract the `maxHeight` in pixel number from parent's relative height. 

This PR alsofixes https://github.com/html-next/vertical-collection/issues/127

@runspired @pzuraq

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,7 @@ module.exports = {
   rules: {
     'quotes': ['error', 'single', {  "allowTemplateLiterals": true, "avoidEscape": true }],
     'ember-suave/no-const-outside-module-scope': 0,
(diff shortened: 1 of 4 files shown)


pzuraq:
I think this would be best pulled out as a helper function similar to [estimateElementHeight](https://github.com/html-next/vertical-collection/blob/master/addon/-private/utils/element/estimate-element-height.js), it could probably use a lot of the same logic

pzuraq:
Failures are dumb, we should remove that ESLint rule

PR #130, opened 2017-09-05
(bugfix): Fix incorrect occluded content height.

billy-addepar:
Sometimes while scrolling a list in a table, I see some open empty space at the top and bottom of the list (see image). After opening the inspection tab, I see height of `occluded-content` is non-zero even though there is no item before/after the list. 

Debugging in the code reveals that the occluded content style height is set to random number when `renderedTotalBefore` is negative. 

https://github.com/html-next/vertical-collection/blob/master/addon/-private/data-view/radar/radar.js#L345 

Since `renderedTotalBefore` is a computed floating number, it's possible that it could be close to 0 and negative. This PR fixes that by using 0 value when `renderedTotalBefore` is negative.

screen shot 2017-09-05 at 11 28 09 am copy

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -342,10 +342,10 @@ export default class Radar {
     const afterItemsText = totalItemsAfter === 1 ? 'item' : 'items';
 
     // Set padding element heights, unset itemContainer's minHeight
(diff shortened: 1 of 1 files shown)


billy-addepar:
@pzuraq @runspired

PR #134, opened 2017-09-16
[BREAKING] make default `containerSelector="body"`

runspired:
This PR changes the default behavior of the collection from expecting the immediate parent div to have scrollable styles defined to expecting that the user intends to rely on the scrolling of the viewport.

To restore the previous default behavior for the collection, set `containerSelector="*"`.

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -44,17 +44,15 @@ const VerticalCollection = Component.extend({
   shouldRecycle: true,
 
   /*
-   * A selector string that will select the element from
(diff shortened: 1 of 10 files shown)


runspired:
@g-cassie thanks for weighing in! To be quite honest I've often wondered whether more folks use body scrolling or element scrolling. In the wild, I tend to encounter body scrolling the most, and I've generally relied on it more myself as well (unless looking to eliminate the safari scroll bounce on mobile).

I'd be very interested in there being more discussion on this, ultimately I want the default to be the thing that folks expect naturally.

runspired:
Closing in favor of #149

PR #140, opened 2017-09-21
(bugfix): condition check on teardown

### LINKED ISSUE #139: Remove scroll handlers can be called before they are added
If teardown occurs before the first scheduled sync, then `removeScrollHandler` will error because the scroll handler hasn't been installed yet.

This can happen if a user navigates away from the page that displays the component at precisely the moment between init and sync.

ming-codes:
Fixes #139

@pzuraq Please review

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -205,8 +205,12 @@ const VerticalCollection = Component.extend({
     this._radar.destroy();
     clearTimeout(this._nextSendActions);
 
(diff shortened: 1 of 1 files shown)


pzuraq:
Thanks for the fix! Will cut a new release soon, by the end of the week

anehx:
When will this change be released?

pzuraq:
@runspired is going to review the two open PRs by this afternoon, then we will push beta.9

PR #142, opened 2017-10-03
Docs for Vertical Collection Component & README

alexander-alvarez:
Please lmk where you think improvements could be made.

Also I need help with : https://github.com/html-next/vertical-collection/pull/142/files#diff-dd82fd2ab5100320f9dd5248149349efR76

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -32,7 +32,7 @@ const VerticalCollection = Component.extend({
   /**
    * Property name used for storing references to each item in items. Accessing this attribute for each item
    * should yield a unique result for every item in the list.
-   * 
+   *
    * @property key
    * @type String
    * @default '@identity'
@@ -63,8 +63,10 @@ const VerticalCollection = Component.extend({
 
   // –––––––––––––– Optional Settings
   /**
-   * Indicates if this component's height will change after initial render.
-   * If true, item tracking will be done using the StaticRadar, otherwise the DynamicRadar.
+   * Indicates if the occluded items' heights will change or not.
+   * If true, the vertical-collection will assume that items' heights are always equal to estimateHeight;
+   * this is more performant, but less flexible.
+   *
    * @property staticHeight
    * @type Boolean
    */
@@ -73,7 +75,13 @@ const VerticalCollection = Component.extend({
   /**
    * Indicates whether or not list items in the Radar should be reused on update of virtual components (e.g. scroll).
    * This yields performance benefits because it is not necessary to repopulate the component pool of the radar.
-   * // WHEN WOULD WE WANT TO SET THIS TO FALSE?
+   * Set to false when recycling a component instance has undesirable ramifications including:
+   *  - When using `unbound` in a component or sub-component
+   *  - When using init for instance state that differs between instances of a component or sub-component
+   *      (can move to didInitAttrs to fix this)
+   *  - When templates for individual items vary widely or are based on conditionals that are likely to change
+   *      (i.e. would defeat any benefits of DOM recycling anyway)
+   *
    * @property shouldRecycle
    * @type Boolean
    */
@@ -94,8 +102,9 @@ const VerticalCollection = Component.extend({
 
   // –––––––––––––– Performance Tuning
   /**
-   * The amount of extra room in pixels to keep visible and invisible on
-   * either side of the viewport. This value needs to be greater than 0.
+   * The amount of extra items to keep visible on either side of the viewport -- must be greater than 0.
+   * Increasing this value is useful when doing infinite scrolling and loading data from a remote service,
+   * with the desire to allow records to show as the user scrolls and the backend API takes time to respond.
    *
    * @property bufferSize
    * @type Number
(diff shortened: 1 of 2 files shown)


addon/components/vertical-collection/component.js
```diff
@@ -29,18 +29,54 @@ const VerticalCollection = Component.extend({
...
+   * @property items
+   * @type Array
+   * @required
+   */
   items: null,
 
   // –––––––––––––– Optional Settings
+  /**
+   * Indicates if this component's height will change after initial render.
+   * If true, item tracking will be done using the StaticRadar, otherwise the DynamicRadar.
```
runspired: @pzuraq confirm that better wording would be `Indicates that this component's height might change after initial render or may differ from 'estimateHeight'`

addon/components/vertical-collection/component.js
```diff
@@ -29,18 +29,54 @@ const VerticalCollection = Component.extend({
...
+   * @property staticHeight
+   * @type Boolean
+   */
   staticHeight: false,
 
+  /**
+   * Indicates whether or not list items in the Radar should be reused on update of virtual components (e.g. scroll).
+   * This yields performance benefits because it is not necessary to repopulate the component pool of the radar.
+   * // WHEN WOULD WE WANT TO SET THIS TO FALSE?
```
runspired: Set to `false` when recycling a component instance has undesirable ramifications including:

- use of `unbound` by the component or a sub-component
- use of `init` for instance state by the component or a sub-component that differs between instances (can move to `didInitAttrs` to fix this)
- the templates for individual items vary widely or are based on conditionals that are likely to change (ie would defeat any benefits of DOM recycling anyway)

addon/components/vertical-collection/component.js
```diff
@@ -57,33 +93,48 @@ const VerticalCollection = Component.extend({
   containerSelector: null,
 
   // –––––––––––––– Performance Tuning
-  /*
-   * how much extra room to keep visible and invisible on
-   * either side of the viewport.
+  /**
+   * The amount of extra room in pixels to keep visible and invisible on
+   * either side of the viewport. This value needs to be greater than 0.
```
runspired: this used to be `pixels` but it is now a strict Item count I believe cc @pzuraq

addon/components/vertical-collection/component.js
```diff
@@ -57,33 +93,48 @@ const VerticalCollection = Component.extend({
...
    */
   renderFromLast: false,
 
-  /*
-   * If set, the collection will render all of the items passed in
+  /**
+   * If set to true, the collection will render all of the items passed into the component.
+   * This counteracts the performance benefits of using vertical collection, but allows for improved accessibility,
+   * since the high performance of vertical-collection is attributed to rendering only the visible elements of a list,
+   * which is not a viable solution in a screen-reader environment, for example.
```
runspired: I might add a note about this being a viable way to render everything in fastboot / other SEO solutions as well as implementing a listener for `meta f` that toggles this to allow the page to be searchable.

addon/components/vertical-collection/component.js
pzuraq: Agree, but the opposite: `Indicates if the occluded item's heights will change or not. If true, the vertical-collection will assume that item height is always equal to estimateHeight. This is more performant, but less flexible.`

addon/components/vertical-collection/component.js
pzuraq: Yep, strict item count

addon/components/vertical-collection/component.js
pzuraq: Also should note, we actually _do_ have viable occlusion for screen readers. It still needs to be tested out, but the `occluded-content` elements on either side of the list will contain text saying `and {n} items before/after`, and should be clickable. Clicking paginates the list, scrolling it to the next `n` items (for however many are rendered). Generic occlusion a11y OOTB 😁

alexander-alvarez: > as well as implementing a listener for meta f that toggles this to allow the page to be searchable.

Can you expand on this? I'm not sure what this means

pzuraq: He means like if you do `cmd-f` to search the webpage, flipping this option would allow users to search the whole text content by rendering all of the content

alexander-alvarez: ahhh... got it.. don't use mac... :smile:

alexander-alvarez:
Second iteration

addon/components/vertical-collection/component.js
alexander-alvarez: > Also should note, we actually do have viable occlusion for screen readers. It still needs to be tested out, but the occluded-content elements on either side of the list will contain text saying and {n} items before/after, and should be clickable. Clicking paginates the list, scrolling it to the next n items (for however many are rendered). Generic occlusion a11y OOTB :grin:
 
I punted on adding this to the docs, but we should definitely include it in an accessibility section once tested :D

pzuraq:
Awesome, thanks again for doing this!

PR #143, opened 2017-10-16
feat(scaling): Makes the collection work with scaling

pzuraq:
When used in a container with `transform: scale`, the collection would break due to measurements being off. For example, we would measure an item that was normally 20px to be scaled by say 10px, and then we would render the measured value, which would be scaled again down to 5px. Lots of small bugs popped up due to this.

This PR adds a `getScaledClientRect` helper which accepts a scale value and element and returns the client rect scaled to the correct size. We determine the scale value in `updateConstants` by comparing the client rect of the scroll container to it's offsetHeight.

This strategy does not account for cases where, for instance, the item container is scaled, but the scroll container isn't. We can probably guard against these cases in debug assertions.

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -3,6 +3,7 @@ import { DEBUG } from '@glimmer/env';
 import Radar from './radar';
 import SkipList from '../skip-list';
 import roundTo from '../utils/round-to';
+import getScaledClientRect from '../../utils/element/get-scaled-client-rect';
 
 export default class DynamicRadar extends Radar {
   constructor(parentToken, options) {
@@ -116,8 +117,10 @@ export default class DynamicRadar extends Radar {
   _measure(measureLimit = null) {
     const {
       orderedComponents,
+      skipList,
+
       _occludedContentBefore,
-      skipList
+      _transformScale
     } = this;
 
     const numToMeasure = measureLimit !== null ? measureLimit : orderedComponents.length;
@@ -132,14 +135,14 @@ export default class DynamicRadar extends Radar {
       const {
         top: currentItemTop,
         height: currentItemHeight
-      } = currentItem.getBoundingClientRect();
+      } = getScaledClientRect(currentItem, _transformScale);
 
       let margin;
 
       if (previousItem !== undefined) {
-        margin = currentItemTop - previousItem.getBoundingClientRect().bottom;
+        margin = currentItemTop - getScaledClientRect(previousItem, _transformScale).bottom;
       } else {
-        margin = currentItemTop - _occludedContentBefore.getBoundingClientRect().bottom;
+        margin = currentItemTop - getScaledClientRect(_occludedContentBefore, _transformScale).bottom;
       }
 
       const itemDelta = skipList.set(itemIndex, roundTo(currentItemHeight + margin));
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -267,11 +267,15 @@ export default class Radar {
     const scrollContainerOffsetHeight = _scrollContainer.offsetHeight;
     const { height: scrollContainerRenderedHeight } = _scrollContainer.getBoundingClientRect();
 
-    // Represents the opposite of the scale, if any, applied to the collection. Check for equality
-    // to guard against floating point errors and divide by zero (both should be zero if one is)
-    const transformScale = scrollContainerOffsetHeight !== scrollContainerRenderedHeight
-      ? scrollContainerOffsetHeight / scrollContainerRenderedHeight
-      : 1;
+    let transformScale;
+
+    // transformScale represents the opposite of the scale, if any, applied to the collection. Check for equality
+    // to guard against floating point errors, and check to make sure we're not dividing by zero (default to scale 1 if so)
+    if (scrollContainerOffsetHeight === scrollContainerRenderedHeight || scrollContainerRenderedHeight === 0) {
+      transformScale = 1;
+    } else {
+      transformScale = scrollContainerOffsetHeight / scrollContainerRenderedHeight;
+    }
 
     const { top: scrollContentTop } = getScaledClientRect(_occludedContentBefore, transformScale);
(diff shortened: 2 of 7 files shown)


addon/-private/data-view/radar/dynamic-radar.js
```diff
@@ -133,14 +136,14 @@ export default class DynamicRadar extends Radar {
...
+      } = getScaledClientRect(currentItem, _transformScale);
 
       let margin;
 
       if (previousItem !== undefined) {
-        margin = currentItemTop - previousItem.getBoundingClientRect().bottom;
+        margin = currentItemTop - getScaledClientRect(previousItem, _transformScale).bottom;
       } else {
-        margin = currentItemTop - itemContainer.getBoundingClientRect().top - totalBefore;
+        margin = currentItemTop - getScaledClientRect(itemContainer, _transformScale).top - totalBefore;
```
runspired: we seem to be duplicating work here

addon/-private/data-view/radar/radar.js
```diff
@@ -36,6 +37,7 @@ export default class Radar {
     this._estimateHeight = 0;
     this._scrollTopOffset = 0;
     this._scrollContainerHeight = 0;
+    this._transformScale = 0;
```
runspired: shouldn't the default be `1` ?

addon/-private/data-view/radar/radar.js
```diff
@@ -207,24 +209,33 @@ export default class Radar {
...
+
+    const { top: itemContainerTop } = getScaledClientRect(itemContainer, transformScale);
+    const { top: scrollContainerTop } = getScaledClientRect(scrollContainer, transformScale);
 
     let maxHeight = 0;
+
     if (scrollContainer instanceof Element) {
-      maxHeight = estimateElementMaxHeight(scrollContainer);
-    }
+      const elementMaxHeight = window.getComputedStyle(scrollContainer).maxHeight;
```
runspired: will this break in fastboot?

addon/-private/utils/element/estimate-element-height.js
```diff
@@ -1,48 +1,31 @@
...
   }
 
   if (fallbackHeight.indexOf('em') !== -1) {
-    return Math.max(getEmHeight(element, fallbackHeight), 1);
+    return getEmHeight(element, fallbackHeight);
   }
 
-  // px or no units
-  return Math.max(parseInt(fallbackHeight, 10), 1);
+  return parseInt(fallbackHeight, 10);
```
runspired: does this fix a bug where we no longer need to ensure a min height of 1? 🎉

addon/-private/utils/element/get-scaled-client-rect.js
```diff
@@ -0,0 +1,10 @@
+export default function getScaledClientRect(element, scale) {
+  const rect = element.getBoundingClientRect();
```
runspired: we should return rect directly if scale === `1` which is the most common case.

addon/-private/data-view/radar/radar.js
```diff
@@ -207,24 +209,33 @@ export default class Radar {
...
-    } = scrollContainer.getBoundingClientRect();
+    // The scroll container's offsetHeight will reflect the actual height of the element, while
+    // it's measured height via bounding client rect will reflect the height with any transformations
+    // applied. We use this to find out the scale of the items so we can store measurements at the
+    // correct heights.
+    const scrollContainerOffsetHeight = scrollContainer.offsetHeight;
+    const { height: scrollContainerRenderedHeight } = scrollContainer.getBoundingClientRect();
+
+    // Represents the opposite of the scale, if any, applied to the collection
+    const transformScale = scrollContainerOffsetHeight > 0 ? scrollContainerOffsetHeight / scrollContainerRenderedHeight : 1;
```
runspired: I'm unsure but this may require a guard against floating point errors

addon/-private/data-view/radar/radar.js
pzuraq: yes, was just defaulting it to a number but `1` would make more sense

addon/-private/utils/element/get-scaled-client-rect.js
pzuraq: good point, will update

addon/-private/data-view/radar/radar.js
pzuraq: Yeah, I thought about that, unsure if rounding would be better or worse in the end. I can try to come up with some tests for it

addon/-private/data-view/radar/radar.js
pzuraq: It should not, we should never get to this function in fastboot. We render solely based on `initialRenderCount` and never schedule anything. I think currently that's based on the fact that RAF doesn't work in fastboot, it should be more explicit. That could actually be a part of the other PR, since it's ultimately about normalizing "starting" the Radar.

addon/-private/utils/element/estimate-element-height.js
pzuraq: ah, actually we still need to define some kind of behavior here, since our pool size is determined by `scrollContainerHeight / estimateHeight`. 

I think we could default `estimateHeight` to `1px` or something, but it would probably be a better idea to assert. Either way, I don't think that belongs in `estimateElementHeight`

addon/-private/data-view/radar/dynamic-radar.js
pzuraq: We won't be, this case will only happen on the very first item. I can add a comment to make this more explicit

addon/-private/data-view/radar/radar.js
pzuraq: Ah, actually it is explicit because we only do that in `didInsertElement`, which does not run in Fastboot

addon/-private/data-view/radar/dynamic-radar.js
runspired: I'm not sure I follow,  it seems like you are generating the boundingClientRect twice

runspired: Nvm I'm dense carry on

PR #144, opened 2017-10-17
feat(tests): Adds a test waiter for the RAF scheduler

pzuraq:
Adds a test waiter that polls until the scheduler's job queue is empty.

The reason for using the `registerWaiter` API despite it's warts is ultimately that it really is the only way to conveniently integrate with the `wait` helper. This has a few implications:

* It's the easiest way to support legacy test helpers (globals)
* It's the only way to completely prevent the user from having to think about whether an action may-or-may-not schedule a RAF. Ex:

```js
test('test a scrolling thing', async function() {
  scrollTo('.scroll-container', 0, 100);

  // ok, that definitely schedules a RAF, better wait
  await waitForRender();

  click('.button-that-causes-scroll')

  // that'll probably cause a RAF, better wait
  await waitForRender();

  fillIn('.search', 'query');

  // that... might cause a RAF? Better wait
  await waitForRender();
});
```

We end up wanting to wait for most things anyways, or at the least not having to worry about waiting.

### Alternatives

* Provide a `waitForRender` helper. The bad part about this would be the mental overhead developers would have to add whenever touching anything that could cause a RAF. Plus, could be tricky to integrate with legacy globals helpers.
* `ember-native-dom-helpers` allows you to provide an alternative `wait` function by setting it on `window.wait`. This seems just as funky, plus only works for `ember-native-dom-helpers`, but we could recommend that user use that instead.

Documentation forthcoming, there will be some larger PRs coming up both here and in the smoke-and-mirrors repo to update all the docs for v1.0


addon/-private/data-view/radar/radar.js
```diff
@@ -471,6 +471,7 @@ export default class Radar {
       const offset = this.getOffsetForIndex(newFirstItemIndex);
 
       this.scrollContainer.scrollTop = offset + this._scrollTopOffset;
+      this.scheduleUpdate();
```
runspired: this looks like a bugfix?

addon/-private/data-view/radar/radar.js
```diff
@@ -487,6 +488,7 @@ export default class Radar {
...
```
runspired: same as above, curious what the issue discovered was.

addon/-private/data-view/radar/radar.js
runspired: 👍

PR #149, opened 2017-10-19
[BREAKING] refactor(tag): Makes vertical-collection tagless

pzuraq:
This PR makes vertical-collection tagless by default. This is inline with the best-effort non-interventionist policy that Smoke and Mirrors has - by removing the base tag, the collection is effectively the same as an `{{each}}` in terms of structure, and has virtually no effect on the users styles.

Currently there are two common issues caused by the component having a tag:

1. Percent styles effectively can't be used for elements, as they would always be relative to the `vertical-collection` tag. This would limit the ability for users to make responsive layouts using the collection. Other styles could be affected by the presence of a the tag as well, but this is definitely the most problematic.

2. The element interferes in various layouts, such as in `table` elements where users would have to know to set the `tagName="tbody"` to make everything work.

I believe this will prevent future issues as well by preemptively reducing the impact we have on user's styles/structure.

An additional benefit of this refactor is it cleans up the flow for starting the Radar, and clearly defines ownership of the scrolling logic, which had been split between the component and Radar for quite some time. The way Radar now finds the scroll container is via the `occluded-content` elements which it adds to the DOM via `unbound`, effectively making it private state of the Radar. 

Note: Users can still use `tagName` and everything should work as expected, collections are tagless by default but having a tag shouldn't break anything.

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -9,6 +9,14 @@ import insertRangeBefore from '../utils/insert-range-before';
 import objectAt from '../utils/object-at';
 import roundTo from '../utils/round-to';
 
+import {
+  addScrollHandler,
+  removeScrollHandler
+} from '../utils/scroll-handler';
+
+import Container from '../container';
+
+import closestElement from '../../utils/element/closest';
 import { estimateElementHeight, estimateElementMaxHeight } from '../../utils/element/estimate-element-height';
 
 const {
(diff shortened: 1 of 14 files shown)


pzuraq:
Also, supercedes #134 by making `containerSelector="*"` by default

addon/-private/data-view/radar/radar.js
```diff
@@ -460,60 +532,46 @@ export default class Radar {
...
 
-      this.scrollContainer.scrollTop = offset + this._scrollTopOffset;
+      this._scrollContainer.scrollTop = offset + this._scrollTopOffset;
       this.scheduleUpdate();
     }
   }
 
-  // If `renderAll` is true, render components for all items. We intercept this here because
-  // for all other behavior (action sending) we want to maintain the "correct" item indexes
-  get renderedFirstItemIndex() {
```
pzuraq: These methods were originally convenience methods, but it felt like they were obfuscating what was actually going on here. We've relied too much on getters/setters in the past, it felt like it would be better to inline the logic since it's only really used in one place.

addon/-private/data-view/radar/radar.js
```diff
@@ -255,20 +306,43 @@ export default class Radar {
...
-      renderedLastItemIndex,
-      renderedTotalBefore,
-      renderedTotalAfter,
-      totalRendered
+      totalItems
     } = this;
 
+    let renderedFirstItemIndex, renderedLastItemIndex, renderedTotalBefore, renderedTotalAfter;
+
+    if (renderAll === true) {
```
pzuraq: Side-effect of this change is that `renderAll` will supercede `initialRenderCount`, which I think makes sense. The order could be reversed here though, meaning that `initialRenderCount` would always win and `renderAll` would only apply after the Radar has started.

addon/components/vertical-collection/component.js
```diff
@@ -230,51 +226,15 @@ const VerticalCollection = Component.extend({
...
-        this._lastEarthquake = top;
-      }
-    };
-
-    this._resizeHandler = () => {
-      this._radar.scheduleUpdate();
-    };
-
-    addScrollHandler(this._scrollContainer, this._scrollHandler);
-    Container.addEventListener('resize', this._resizeHandler);
```
g-cassie: @pzuraq I think we have a bug related to this removal

basically we have responsive elements in the list that get shorter when a break point is hit. As a result in certain cases after resizing there are not enough rendered items to fill the scroll container. It fixes once you scroll.

I read through the PR but don't see any mention of this removal.  Is there a path forward to add our own resize handler?

runspired: I found a path forward the other night.

Cc @pzuraq

g-cassie: Ok, let me know if I can help. We are going to revert to `beta.8` for now.

PR #151, opened 2017-10-23
feat(dynamic-pool): Implements a dynamic component pool with incremental render

pzuraq:
This PR returns vertical-collection to a dynamic component pool, adding and removing components from the DOM as necessary based on actual component measurements. The goal is that we never render more items than we need to, allowing the collection to respond to varying sizes of items (e.g. items that are much larger or smaller than the estimate height).

The last time this strategy was attempted it was reverted due to being a performance regression. Destroying and re-creating components dynamically cost too much compared to the benefits. This time we're getting around that issue by never actually destroying any components. We simply remove their DOM, appending it to a temporary container element and then re-appending it when we need another component.

An additional feature required to make this fully work is incremental render. If enabled, the collection will measure and schedule render if it turned out it was wrong about the number of items needed,
for instance because they were too small.

TODO:
- [x] Add tests for incremental render
- [x] Better tests for:
  - [x] Prepend measurements (silent night)
  - [x] `renderFromLast`
  - [x] `didEarthquake`

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -43,13 +43,15 @@ export default class DynamicRadar extends Radar {
 
   afterUpdate() {
     // Schedule a check to see if we should rerender
-    this._nextIncrementalRender = this.schedule('sync', () => {
-      this._nextIncrementalRender = null;
+    if (this._nextIncrementalRender === null && this._nextUpdate === null) {
+      this._nextIncrementalRender = this.schedule('sync', () => {
+        this._nextIncrementalRender = null;
 
-      if (this._shouldScheduleRerender()) {
-        this.update();
-      }
-    });
+        if (this._shouldScheduleRerender()) {
+          this.update();
+        }
+      });
+    }
 
     super.afterUpdate();
   }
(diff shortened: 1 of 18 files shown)


gwak:
It seems that this commit created a regression in how the scrollTop of the container element is set in afterUpdate of the radar.js file. If the scroll container is scrolled, it will be reset to 0 because the difference is equal to the scrollTop, which in my use case is wrong. Each time I hide/show my vertical-collection which is in the middle of the container panel, the scroll goes back to the top.

```javascript
if (scrollDiff !== 0) {
this._scrollContainer.scrollTop += scrollDiff;
} 
```

pzuraq:
So you're saying that you have a layout with a collection that is in the middle of the scroll container, offset by some amount, and when you render it it resets the scroll position to the top? Is it in a conditional of some kind?

gwak:
Yes exactly, my vertical-collection is in a collapsible panel, so every time it's rendered,  the conatiner scollTop is reset to 0.

pzuraq:
Sounds like that shouldn't be happening, I agree. Can you possibly PR a failing test or post a reproduction so we can nail down the exact issue? I can try to fix it after the holidays, but the best way to do it is with a reproduction case to make sure it works.

gwak:
Hi,
Thanks for considering my use-case. Here's a PR with a failing test : 
https://github.com/html-next/vertical-collection/pull/164
Thanks

PR #165, opened 2018-01-03
Update README.md

xomaczar:


xomaczar:
Guess so, sorry did not notice it

pzuraq:
Thanks for the fix!

PR #168, opened 2018-01-18
bugfix(timing): Pulls prepend/append logic into Radar

pzuraq:
Pulls the prepend/append/reset logic into Radar due to issues with
timing guarantees. This guaratees that we will check the state of the
array when we are going to update, not when we are scheduling the
change. In certain edge cases, updates to the array can occur between
when the computed property is calculated and the Radar actually updates.

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -8,6 +8,7 @@ import VirtualComponent from '../virtual-component';
 import insertRangeBefore from '../utils/insert-range-before';
 import objectAt from '../utils/object-at';
 import roundTo from '../utils/round-to';
+import { isPrepend, isAppend } from '../utils/mutation-checkers';
 
 import {
   addScrollHandler,
@@ -19,6 +20,7 @@ import Container from '../container';
 import closestElement from '../../utils/element/closest';
 import estimateElementHeight from '../../utils/element/estimate-element-height';
 import getScaledClientRect from '../../utils/element/get-scaled-client-rect';
+import keyForItem from '../../ember-internals/key-for-item';
 
 const {
   A,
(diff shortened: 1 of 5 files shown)


pzuraq:
Supercedes #147

addon/-private/data-view/radar/radar.js
```diff
@@ -224,7 +244,7 @@ export default class Radar {
   }
 
   afterUpdate() {
-    const { totalItems } = this;
+    const { _prevTotalItems: totalItems } = this;
```
pzuraq: @runspired I did this to counter a very strange heisenbug that occurs in the scroll tests. Occasionally, somehow, `totalItems` will _change_ before this function is run. I believe this is only happening in PromiseArray scenarios, which makes sense, but I can't figure out _how_ - everything between the beginning on this update and now should all be run synchronously. How is an array being updated during a synchronous function sequence, with no breaks at all for async functions?

addon/-private/data-view/radar/radar.js
```diff
@@ -215,7 +226,16 @@ export default class Radar {
     });
   }
 
+  updateItems() {
```
pzuraq: This is a real pain, but basically we only want to do a `reset/prepend/append` when the items have changed, but we don't want to run or cache the final indexes _until_ the RAF because the array can change in the interim (especially in the case of PromiseArrays)

addon/-private/data-view/radar/radar.js
pzuraq: Looking at it closer, it seems like this is happening because the promise is being resolved on the runloop, which makes sense. Seems strange that it would resolve on a nested runloop, but it seems to be happening

runspired: 🤔

runspired:
👍

PR #181, opened 2018-02-05
Remove "find host" loop

twokul:
"find host" loop breaks import functionality with latest versions of Ember CLI. [It is a part of `app.import`](https://github.com/ember-cli/ember-cli/blob/v3.0.0-beta.2/lib/models/addon.js#L675) now and no longer needed here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -125,15 +125,6 @@ module.exports = {
     this._super.included.apply(this, arguments);
     this.checker = new VersionChecker(app);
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@twokul did you check that this still works properly as a nested addon after this change? iirc it has to be `this.import()` in this case and `app` also seems to be used for other things that might break when it is referring to the parent addon instead. 

We could also use `findHost()` though we might have to polyfill it depending on the Ember CLI version that is used.

PR #191, opened 2018-03-02
Bugfix - Fix issue where IE crashed trying to set source of a null reference

### LINKED ISSUE #74: Vertical collection does not work very well with IE11 and Edge

localpcguy:
Fixes #74
Caused Error: "TypeError: Unable to get property 'source' of undefined or null reference" (in meta.js)

As discussed in the issue, changing the code to only set `this.__ember_meta__` when in an older version of ember is found. Also changed it to `undefined` instead of `null` so the comparison in Ember meta.js file will catch it when it is unset.


pzuraq:
Thanks for the PR!

PR #192, opened 2018-03-10
Add fastboot smoke test

rwwagner90:
This should give us a smoke test for fastboot, but it passes somehow, which I did not expect, since this same thing fails in my app.


rwwagner90:
Closing this. If someone wants to pick it back up, feel free!

PR #194, opened 2018-03-13
API to specify custom occluded content tag

billy-addepar:
Allow user to specify custom occluded tag through `occlusionTagName` field. This allows component like table to comply with table semantics.


billy-addepar:
@runspired @pzuraq

PR #196, opened 2018-03-19
[WIP] generify radar dimension

kybishop:
Still a few more things to do

### DIFF
--- a/addon/-private/data-view/radar/dynamic-radar.js
+++ b/addon/-private/data-view/radar/dynamic-radar.js
@@ -60,15 +60,15 @@ export default class DynamicRadar extends Radar {
   _updateConstants() {
     super._updateConstants();
 
-    if (this._calculatedEstimateHeight < this._minHeight) {
-      this._minHeight = this._calculatedEstimateHeight;
+    if (this._calculatedEstimateSize < this._minHeight) {
+      this._minHeight = this._calculatedEstimateSize;
     }
 
-    // Create the SkipList only after the estimateHeight has been calculated the first time
+    // Create the SkipList only after the estimateSize has been calculated the first time
     if (this.skipList === null) {
-      this.skipList = new SkipList(this.totalItems, this._calculatedEstimateHeight);
+      this.skipList = new SkipList(this.totalItems, this._calculatedEstimateSize);
     } else {
-      this.skipList.defaultValue = this._calculatedEstimateHeight;
+      this.skipList.defaultValue = this._calculatedEstimateSize;
     }
   }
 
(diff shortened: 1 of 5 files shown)


addon/-private/data-view/radar/dynamic-radar.js
```diff
@@ -60,15 +59,15 @@ export default class DynamicRadar extends Radar {
   _updateConstants() {
     super._updateConstants();
 
-    if (this._calculatedEstimateHeight < this._minHeight) {
```
runspired: seems odd for _minHeight to still be here in a generalized version

pzuraq:
I'm generally onboard with this at a high level. Agree that automated perf tests would be good to add but given that @kybishop is going to need `horizontal-collection` sooner rather than later, I think we could probably do with manual tests based on `ember-macro-benchmark`. Do you think you could lay out a plan for automated performance tests if we're going to block on doing this work until we get them?

Also agree that the current strategy of branching and adding measurement functions feels a bit off, though I'm unsure of what a better strategy might be. I could see a functional approach, though it would introduce a fair amount of boilerplate (constantly having to pass in `dimension`) but it would probably isolate exactly what we are measuring pretty well, so I'd lean that direction as well.

addon/-private/data-view/radar/dynamic-radar.js
kybishop: Still a WIP, will get to this soon.

kybishop:
Color me surprised if a functional approach is more performant than set-and-forget on init, though v8 optimization _is_ largely a black box to me...

I'm very much on board with some perf tests. Would be great to stress test both approaches.

addon/-private/utils/element/estimate-element-size.js
```diff
@@ -0,0 +1,34 @@
...
+  // We use offsetHeight/offsetWidth here to get the element's true size, rather than the
+  // bounding rect which may be scaled with transforms
+  let parentSize = dimension === 'height' ? element.offsetHeight : element.offsetWidth;
+  let percent = parseFloat(fallbackSize);
+
+  return (percent * parentSize / 100.0);
+}
+
+// TODO(kjb) How do we handle this for width? font-size only refers to font height, though we could
+// make a rough estimate of width based on some general-case constant.
```
kybishop: This should be addressed before moving forward with horizontal collection. It's the only thing I'm somewhat hung-up on since there is no definitive answer.

runspired: font-size applied to width as much as to height

kybishop: Maybe this is a stupid question, but then why are there monospaced fonts? It's my understanding that the width of a font is not directly correlated to its height, but arbitrary.

pzuraq: I think `em` and `rem` font-size is a separate concept from actual rendered font size. If you were to say that `em` was based on the current font-size of `12px`, and then to say that this item was `width: 2em`, it would be 24px.

rwwagner90:
@pzuraq @kybishop where did we leave off on this?

kybishop:
I've been away from Ember for over half a year now. Someone is welcome to take up the mantel on this if they want to, but I definitely don't have the time to jump back into Ember stuff these days.

PR #233, opened 2018-10-02
[Deprecation fix] `this.sendAction` will be deprecated

raghuvaran:
[RFC](https://github.com/emberjs/rfcs/blob/master/text/0335-deprecate-send-action.md)

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -155,7 +155,7 @@ const VerticalCollection = Component.extend({
   isEmpty: empty('items'),
   shouldYieldToInverse: readOnly('isEmpty'),
 
-  virtualComponents: computed('items.[]', 'renderAll', 'estimateHeight', 'bufferSize', function () {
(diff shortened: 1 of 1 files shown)


pzuraq:
`vertical-collection@v1.0.0` will continue to support Ember v1, so we cannot remove `this.sendAction` for the time being. When we do get v1.0.0 out the door, we will pivot to creating v2.0.0 which will be a rewrite that drops support for 1.11 and allows us to do this.

raghuvaran:
Ok thanks!

raghuvaran:
I think the tests are using `Ember.evented` style for testing `action` invocation. Since `action` here is not a closure function, tests are timing out.

https://github.com/html-next/vertical-collection/blob/b76811e2bdfa0d947b55b343872a35379744a3a5/tests/integration/scroll-test.js#L52-L77

efx:
It looks like this was already fixed in #248. Can we close this PR?
https://github.com/html-next/vertical-collection/commit/e2b6f83cb0d2e0f07e3ffe60c124657f886b9196

PR #244, opened 2018-11-12
Fix sendAction() deprecation

fpauser:
...by calling lastReached, firstReached, lastVisibleChanged and firstVisibleChanged actions directly.


pzuraq:
Hey @fpauser, there is actually another PR open that’s attempting to do this as well. The issue is that VC’s support matrix for v1 includes versions of Ember which do not support closure actions.

It should be possible to use`ember-compatibility-helpers` To detect support for closure actions, which is what the other PR does, but then we have to update our test suite which only uses string actions. Happy to provide guidance for that!

efx:
This is the other PR for anyone interested: https://github.com/html-next/vertical-collection/pull/233

PR #253, opened 2019-04-02
chore: use Node 8 as default

cyril-sf:
Node 6 is EOL in April 2019


cyril-sf:
@runspired @pzuraq It looks like the build is passing. It will also fix the broken build in master.

pzuraq:
👍

PR #257, opened 2019-04-23
[CHORE] Fix the build

cyril-sf:
- Use assign from ember-assign-polyfill instead of Ember.merge
This will remove many deprecation warnings raised in the test environment. It seems like it makes Travis fail to read the result of the tests otherwise.

Screen Shot 2019-04-23 at 8 51 44 AM

- Rename index to itemIndex in `number-slide`
I haven't tracked down the problem but it seems like Ember thinks that
`{{index}}` inside the `number-slide` component is not the property but the template
for the `index` route.
I doubt this is a bug with `vertical-collection`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
   "devDependencies": {
     "bootstrap": "~3.3.5",
     "broccoli-asset-rev": "^3.0.0",
+    "ember-assign-polyfill": "^2.6.0",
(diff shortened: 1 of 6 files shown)


cyril-sf:
@pzuraq @runspired Can you take a look?

pzuraq:
Thanks for tracking these down man! Really appreciate it 😄

PR #258, opened 2019-04-23
Chore: Use Babel 7

### LINKED ISSUE #229: Support ember-cli-babel 7
When trying to use with `ember-cli-babel` `7.*` I see the following build error:

```
Build Error (broccoli-persistent-filter:Babel > [Babel: Modules for Templates]) in @html-next/vertical-collection/components/vertical-collection/template.js

Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel. (While processing preset: "/some/directories/here/node_modules/@babel/preset-env/lib/index.js")
```

h2blake:
Closes #229


index.js
```diff
@@ -108,17 +94,9 @@ module.exports = {
     const opts = {
       loose: true,
       plugins,
-      postTransformPlugins: [StripClassCallCheckPlugin],
-      exclude: [
-        'transform-es2015-block-scoping',
-        'transform-es2015-typeof-symbol'
```
h2blake: do we need `@babel/plugin-transform-typeof-symbol` here?

pzuraq:
This will probably be need a rebase on top of the work done by @cyril-sf to fix the build, could help solve these problems

rwwagner90:
@h2blake would you mind resolving the conflicts on this please?

PR #282, opened 2019-08-04
Remove custom rollup and babel stuff

### LINKED ISSUE #229: Support ember-cli-babel 7

rwwagner90:
I'm working to remove some of the extra stuff on top of this addon, so that it is easier for contributors to grok, and gets us closer to a default addon setup.

This is the first thing I am looking to remove. @runspired any issues with this?

Closes #229 
Closes #258

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,24 +1,7 @@
 'use strict';
 
-const StripClassCallCheckPlugin = require('babel6-plugin-strip-class-callcheck');
 const Funnel = require('broccoli-funnel');
-const Rollup = require('broccoli-rollup');
-const merge = require('broccoli-merge-trees');
 const VersionChecker = require('ember-cli-version-checker');
-
-const path = require('path');
-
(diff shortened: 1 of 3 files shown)


rwwagner90:
@runspired is opposed to this, so closing.

PR #289, opened 2019-08-16
Remove compatibility helpers, test only Ember 3.4+

rwwagner90:


.travis.yml
```diff
@@ -49,14 +49,7 @@ jobs:
...
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-1.12
-    - env: EMBER_TRY_SCENARIO=ember-1.13
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.4
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.8
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.12
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.16
-    - env: EMBER_TRY_SCENARIO=ember-lts-2.18
-    - env: EMBER_TRY_SCENARIO=ember-lts-3.4
+      env: EMBER_TRY_SCENARIO=ember-lts-3.4
```
pzuraq: We'll only be explicitly supporting 3.8 in the next release, so I think it would be fine to drop this now and just keep 3.8

Issue #288, opened 2019-08-16
Changelog


amk221:
I'm the process of upgrading my addons. `yarn outdated` suggests I can upgrade vertical-collection from `1.0.0-beta.13` to `1.0.0`

But there are no details in the changelog or on the releases page.

Would be great to get some details, ty!

rwwagner90:
@amk221 not really any changes, just shipping what you were using in the betas as a major version

amk221:
Ok, thank you! I figured that would be all. (Still nice to have the go-to places have the up to date info)

rwwagner90:
Totally agree. We need to start generating changelogs with each release.

PR #293, opened 2019-08-29
automate changelog generation

efx:
Use https://github.com/rwjblue/create-rwjblue-release-it-setup for a
release-it setup.
Addressing https://github.com/html-next/vertical-collection/issues/288#issuecomment-521998860
I've not run the changelog generation yet. Once agreed upon and merged we could do that in a separate PR.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,65 @@
+# Release
+
+Releases are mostly automated using
+[release-it](https://github.com/release-it/release-it/) and
+[lerna-changelog](https://github.com/lerna/lerna-changelog/).
+
+
+## Preparation
+
+Since the majority of the actual release process is automated, the primary
+remaining task prior to releasing is confirming that all pull requests that
+have been merged since the last release have been labeled with the appropriate
+`lerna-changelog` labels and the titles have been updated to ensure they
+represent something that would make sense to our users. Some great information
+on why this is important can be found at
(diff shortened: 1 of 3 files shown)


rwwagner90:
@efx I would prefer github_changelog_generator or something that uses tags on issues and PRs to generate the changelog.

efx:
Sure. lerna-changelog does use tags on the PRs so I'm not sure I understand the objection. You can see it at work here: https://github.com/adopted-ember-addons/ember-pikaday

rwwagner90:
@efx I didn't see anything about lerna here. I'm unfamiliar with release-it but when I quickly read their changelog section, I didn't see anything about using lerna or tags.

efx:
@rwwagner90  no worries; I think part of its name (`lerna`) gives the wrong impression. I chose it because of recommendations in adopted addons discussion and I found it useful to generate the changelog based on labels. I also chose it because of the automated release it setup. 
I'm totally fine with whatever direction you want to head; contributing-wise I opted for the laziest approach I know.

rwwagner90:
@efx if this uses tags, I am not opposed to it. @pzuraq what is used for the other Ember projects for changelogs? Is it still a custom script?

efx:
Woot! I'm opting for whatever is standard across ember projects and is the path of least resistance.

rwwagner90:
Not sure there is a defined standard. I think Ember uses some custom scripts. I use github_changelog_generator for Ember Inspector and other repos.

Issue #296, opened 2019-09-05
shouldRecycle=false breaks UI when scrolling the elements fast.


ahamedalthaf:
'shouldRecycle: false' is breaking the UI when we try to scroll faster. This breaks the UI and completely block the user from doing any activity. 

Please find the error below in the browser console.

**Assertion Failed: value must non-negative**

rwwagner90:
@ahamedalthaf can you try the latest version please?

ahamedalthaf:
@rwwagner90 tried with latest version and still seeing the same issue.

NullVoxPopuli:
I'm experiencing this on emberclear, too :(

ahamedalthaf:
@rwwagner90 I have fixed this issue in the PR raised. Please check from your side.

rwwagner90:
Thanks @ahamedalthaf!

knownasilya:
Was seeing `nextSibling` of undefined in the glimmer vm with this flag as well.

jbottigliero:
While the referenced PR has been merged for some time (looks like it would have landed in `3.0.0`), I'm encountering a similar error in newer versions.

I've tested in both `4.0.1` and `3.1.0`, both versions result in a similar error, the below report is from `4.0.1` (latest at time of writing):

```js
Uncaught Error: Assertion Failed: value must non-negative
    at Object.assert (index.js:181:1)
    at SkipList.set (-private.js:1486:1)
    at DynamicRadar._measure (-private.js:1762:1)
    at DynamicRadar._updateIndexes (-private.js:1656:1)
    at DynamicRadar.update (-private.js:782:1)
    at -private.js:769:1
    at Array.execJob (index.js:36:1)
    at Scheduler.flush (index.js:96:1)
    at index.js:82:1
````

I'll try and follow-up with more details as I begin to investigate...

wozny1989:
Anyone got an idea how to fix this? We have the same reason as @NullVoxPopuli to use `shouldRecycle={{true}}`

NullVoxPopuli:
since I posted my comment, I've found that using derived data _for everything_ solves most problems.

did-insert/did-update/will-destroy are problematic for more than just this scenario, but also outside of Ember. `useEffect` in React suffers from the same problem, as do other effect implementations in other frameworks, such as Preact, Solid, etc.

By using derived data, you appropriately get the correct behavior when you need it (when the root state changes, which is what happens during component re-use)

PR #299, opened 2019-09-21
shouldRecycle=false breaks UI when scrolling the elements fast. Issue#296

ahamedalthaf:
Commit for the below issue.

- shouldRecycle=false breaks UI when scrolling the elements fast.

Issue#296

Reference: https://github.com/html-next/vertical-collection/issues/296

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -556,6 +556,7 @@ export default class Radar {
       virtualComponents,
       _occludedContentAfter,
       _appendComponentPool,
+      shouldRecycle,
       _itemContainer
     } = this;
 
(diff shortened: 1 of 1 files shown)


rwwagner90:
@pzuraq I'm not familiar with the internals. Does this work?

pzuraq:
it's been a long time, I'd have to reboot a lot of context and unfortunately I'm stretched a bit too thin to do that at the moment 😕 I'll try to get back to this when I have time

rwwagner90:
I don't know enough about VC to review this unfortunately.

ahamedalthaf:
@pzuraq Can you please check this once. 
**Root-cause:**
Some components not appended properly when we scroll fast. So we are getting negative values in offset and assertion error is thrown.
**Solution:**
Have created an array 'appendComponentPool' and appending the items after components are rendered. Already we are following the same approach in 'prependComponent' method.

knownasilya:
Also see this, get `nextSibling` errors in the glimmer vm

rwwagner90:
Anyone interested in helping maintain this repo? I don't know enough about the internals myself

pablomaribondo:
@rwwagner90 @pzuraq @runspired Can you guys give this another try ? this error is still happening

runspired (approved):
Overall this change seems good.

I suspect there's a more efficient way to do this in keeping with the original rendering model, as in the original rendering model what this PR brings is exactly how the order of operations would have occurred, but we likely have deviated from that with time.

mixonic:
In `3.0.0-1`

Issue #317, opened 2020-03-06
keyForItem called on null value


Emerson:
We have a mature application that's going through an Ember upgrade and have run into some issues with `vertical-collection`. We have a list of items wrapped in `vertical-collection` that are filtered by a search input. As a user types into an `<input>`, the items in the list are filtered.

In some cases, when the user types quicky and the filtered number of items is zero, we see an error raised by `vertical-collection`:

```
# from ember-metal _get
Cannot call get with 'id' on an undefined object
```

I've tried my best to trace the logic of `vertical-collection` and ended up in the following method:

```js
  _scheduleSendAction(action, index) {
    this._scheduledActions.push([action, index]);

    if (this._nextSendActions === null) {
      this._nextSendActions = setTimeout(() => {
        this._nextSendActions = null;

        run(() => {
          const items = this.get('items');
          const keyPath = this.get('key');
          if (!items || items.length === 0) { return; }

          this._scheduledActions.forEach(([action, index]) => {
            const item = objectAt(items, index);
            /*
            *     ISSUE:
            *
            *     For some reason this method is being run even when `items` is
            *     an empty array. This results in item being `null`, which is
            *     then passed into `keyForItem`, which attempts to get the
            *     id property of a null value, resulting in an error.
            *
            */
            const key = keyForItem(item, keyPath, index);

            // this.sendAction will be deprecated in ember 4.0
            const _action = get(this, action);
            if (typeof _action == 'function') {
              _action(item, index, key);
            } else if (typeof _action === 'string') {
              this.sendAction(action, item, index, key);
            }
          });
          this._scheduledActions.length = 0;
        });
      });
    }
  },
```

andreyfel:
@Emerson we have a similar issue which happens in the tests, but it is quite flaky. I couldn't make it fail consistently. Do you have a failing test for this?

andreyfel:
Here is my take on this issue:
https://github.com/html-next/vertical-collection/pull/324

Emerson:
Interesting - I kind of stopped tracking this issue since there seemed to be no interest in fixing it. We're still getting errors in production related to this - although it doesn't seem to prevent out app from functioning.

We never managed to get a failing test - but we can manually trigger the issue by quickly changing the value in a search field that filters a list managed by vertical collection.

andreyfel:
@Emerson do you think you can try my branch in your app to see if it fixes the issue for you?

PR #318, opened 2020-03-06
Do not run schedule actions if there are no items

Emerson:
Outlined the problem in #317. This is a simple fix that has resolved the issue for me and still passes all tests. Happy to put in a little more work on this, but would need some guidance on where to look 👀.

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -181,6 +181,7 @@ const VerticalCollection = Component.extend({
         run(() => {
           const items = this.get('items');
           const keyPath = this.get('key');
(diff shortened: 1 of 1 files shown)


addon/components/vertical-collection/component.js
```diff
@@ -181,6 +181,7 @@ const VerticalCollection = Component.extend({
         run(() => {
           const items = this.get('items');
           const keyPath = this.get('key');
+          if (!items || items.length === 0) { return; }
```
runspired, line 184: we should also not fire below if index exceeds the bounds of length. Otherwise this LGTM

runspired (approved):

PR #321, opened 2020-08-10
Kick off 2.0-beta

mixonic:
Per https://github.com/Addepar/ember-table/issues/819 we're pursuing a 3.0 release for ember-table which drops IE11 support and Ember pre-2.8. As ember-table is likely the most active contributor of this library, I would also like to release a 2.0 of vertical collection dropping support for IE11 and Ember < 2.8.

This branch should kick that off.


mixonic:
Annoyingly although ember-table supports back to Ember 2.4 (back to 1.11 on master!) this library is only tested in Ember 3.x and does not pass on 2.x.

rwwagner90:
This seems fine so far. If we're dropping IE11, maybe we should also drop Ember 2.x?

mixonic:
Addepar relies on 2.x support both for ember-table and various internal component implementations. Ember-table itself is probably the single largest consumer of this library, and since we can't drop Ember 2.x from ember-table yet we should not drop it from here. In fact I would like to *add* it to the test suite.

Quickly brainstorming, the advantage to dropping 2.x afaict would be a more stable component manager API which may allow an implementation here to operate at a lower level, or at least at a low level on a stable, public API instead of intimate API. But we could probably also use those APIs progressively during a vertical-collection 2.x cycle and drop weight at 3.x.

apellerano-pw:
Is there any plans to integrate the horizontal-collection / virtual-collection pattern proposed in #269 into the next major release?

knownasilya:
I'd love to at least get a Ember 3 and IE11 release, since Ember still supports IE11.

rwwagner90:
I don't really know much about the internals of this repo, but happy to add anyone as maintainers if they want to help get this over the finish line!

mixonic:
In alignment with a proposal at https://github.com/Addepar/ember-table/issues/819#issuecomment-772814116, I've opened a PR at https://github.com/html-next/vertical-collection/pull/325 moving the minimum supported version of Ember for the 2.0 branch to 2.8.

mixonic:
@twokul and I discussed and we think the last item here is the `ember-raf-scheduler` dependency getting a bump.

twokul:
As per discussion, https://github.com/html-next/vertical-collection/pull/328. I believe after it is merged, we can safely merge this PR and proceed with the release of 2.0.

twokul (approved):

.travis.yml
```diff
@@ -22,34 +21,40 @@ env:
...
+    - 2.0-beta
     # npm version tags
     - /^v\d+\.\d+\.\d+/
 
 jobs:
-  fail_fast: true
+  fast_finish: true
   allow_failures:
     - env: EMBER_TRY_SCENARIO=ember-canary
+    - env: EMBER_TRY_SCENARIO=ember-lts-2.8
```
rwwagner90, line 32: Why are we using both Travis and GitHub Actions? Shouldn't we use just one or the other? My preference would be GitHub Actions.

tests/dummy/config/targets.js
```diff
@@ -6,13 +6,6 @@ const browsers = [
   'last 1 Safari versions'
 ];
 
-const isCI = !!process.env.CI;
-const isProduction = process.env.EMBER_ENV === 'production';
```
rwwagner90, line 10: Why remove this code?

rwwagner90 (changes requested):

.travis.yml
mixonic: Oh yeah this file should just get deleted. I thought I had already done it, but I've been bouncing around so many repos maybe I forgot.

tests/dummy/config/targets.js
mixonic: There is no need to build the dummy app with IE9 or any other legacy browser support for testing.

.travis.yml
twokul: Indeed. We use GH Actions as a "regular" CI and Travis to catch the possible issues with "floating" dependencies. I agree we should standardise on one; I would not consider it blocking 2.0. Open to discussing of course! 😄

tests/dummy/config/targets.js
twokul: I do not think it is used anymore? 🤔

rwwagner90: This just comes from ember-cli blueprints though, right? If so, I would like to keep as much in line with the blueprints as we can.

.travis.yml
rwwagner90: I think let's delete Travis for now and go with GH actions. I've got a floating dependencies job setup here if we want to copy that https://github.com/shipshapecode/ember-shepherd/blob/master/.github/workflows/main.yml#L25

twokul: https://github.com/html-next/vertical-collection/pull/329

tests/dummy/config/targets.js
twokul: I am a little confused. We do not want to support IE, afaik. If so, that makes sense to remove the code. Am I missing something? Otherwise, both `isCI` and `isProduction` will be unused 🤔

twokul: https://github.com/html-next/vertical-collection/pull/330

twokul: Deleting it breaks the tests. Sorry, I do not see a good reason to keep this code around. I am happy to follow up with any change later. I do not think we should block the release on this 😄

rwwagner90: @twokul can we please just keep it as it was before this PR? It matched the ember-cli blueprints before.

twokul: I do not understand the reasoning but I have reverted the change as you asked https://github.com/html-next/vertical-collection/pull/321/commits/b0d44dbf9e40d721507321ab124bb5f590947905

twokul: The tests won't pass since `ember-maybe-import-regenerator` is not a dependency anymore

rwwagner90: The reasoning is to match the blueprints for ease of future Ember updates and better diffs. I do not disagree with you that the code is not necessary, I just always want to strive to match the current blueprints in all cases, unless we have a reason to diverge.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+
+  test-ember-try:
+    name: Run Tests
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        ember-version:
+          [
+            # The test suite only support Ember 3.x
+            # ember-lts-2.8,
```
rwwagner90, line 45: Should we uncomment these, if we intend to support 2.8+?

twokul: Now that `.travis.yml` is removed, I think so 🤔

tests/dummy/config/targets.js
mixonic: @rwwagner90 we have a reason to diverge, IE11 is not supported. Right?

.github/workflows/ci.yml
mixonic: There is a comment right above them. The test suite doesn't run in Ember 2.x. The commented lines and comment above them explain that these are disabled due to that flaw.

We should get the tests to pass on 2.x, but I'm don't think we should block it since the CI suite today on master is busted generally.

rwwagner90: It looks like we need to set ember-compatibility-helpers to 1.2.1 https://github.com/pzuraq/ember-compatibility-helpers/issues/47

mixonic: This PR should not be blocked on fixing the 2.x test suite. I'm happy to entertain PRs against the beta or against master after this lands.

The code has copious coverage by ember-table's suite (which is running against 2.x) an against a massive codebase internally at Addepar. If is far more important that we get master into a functional state and get a release than it is to golf out the final legs of test suite repair.

tests/dummy/config/optional-features.json
```diff
@@ -1,3 +1,3 @@
 {
-  "jquery-integration": false
+  "jquery-integration": true
```
mixonic, line 2: Why would you *enable* jquery?

rwwagner90: It's required for Ember 2.x

rwwagner90 (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@html-next/vertical-collection",
-  "version": "1.0.0",
+  "version": "2.0.0-0",
```
rwwagner90, line 3: Should we label this as a beta?

mixonic: This format of version number is commonly used by NPM packages to identify a pre-release. The published package on NPM is tagged `next`, which is basically equivalent to `beta` (only the `latest` tag on npm is for stable packages).

You prefer the `-beta` style versioning? Or were you referring to the npm tag?

mixonic:
I've integrated this into ember-table and into a large codebase. I think we're good to land it and cut 2.0!

package.json
rwwagner90: I'm just used to seeing `-beta` on betas.

PR #324, opened 2020-11-16
Use once instead of setTimeout + run

andreyfel:
In our test environment we were getting an error:
`Cannot call get with 'id' on an undefined object`.
It was quite flaky, happening only occasionally.
However we've noticed that sometimes it is happening in production as well.
There is an open issue in the repo for that bug:
https://github.com/html-next/vertical-collection/issues/317

Since this is a flaky issue it is hard to write a test for it, however
there was one test in our test suite which was flaking more frequently,
so, we took it as reference point to check if the fix is working.
But the test is not easy to isolate and integrate into
vertical-collection repo, so, this commit won't include test as a proof
for the fix.

====

The actual change:

This bug was happening when vertical collection component was trying
to send actions to the users. It was being scheduled through setTimeout
and wrapped in run(). setTimeout() is not bound to the ember runloop,
this is why explicit `run` was needed.

My guess is that because of that the send actions code could be run by
the moment when the item for which the action was meant to be sent
had gone from the vertical collection.

In fact Ember has API to schedule some code to be run within the proper
queue which is the `actions` queue in this particular case. Using that
API seems to be fixing `Cannot call get with 'id' on an undefined object`
exception for us.

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -2,7 +2,7 @@ import { empty, readOnly } from '@ember/object/computed';
 
 import Component from '@ember/component';
 import { get, computed } from '@ember/object';
-import { run } from '@ember/runloop';
+import { cancel, once } from '@ember/runloop';
 import layout from './template';
 
 import { scheduler, Token } from 'ember-raf-scheduler';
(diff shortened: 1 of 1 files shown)


andreyfel:
@pzuraq could you please take a glance at this PR? I would love to get some feedback on it.

andreyfel:
@mixonic @twokul Hey guys! Could you please take a look at this PR?

Emerson:
I'm considering trying this branch out - we continue to have errors related to this issue. It doesn't appear to really impact our customers, but it sure creates a lot of noise in our error logs.

Emerson:
hey @andreyfel - I wanted to chime in and let you know that we published your fork/fix to production and have seen our errors disappear. We have a good number of users with different browsers (IE11 included) and none of them have had any issues since we pushed this out.

andreyfel:
Hi @Emerson! Thank you for that feedback! I'm glad that my fix helped!

mixonic:
Hm, writing a test for this should be possible. It is something like:

1. content is rendered (`items` is full)
2. scroll happens which triggers the actions from radar (`fistVisibleChanged` for example)
3. In an animation frame or promise microtask, change `items` to be empty (or remove the item the even was firing for)
4. assert no errors and that the event fires with the expected arguments

The change here is reasonable, but would cause app-written logic based on the radar events to be run in the same runloop as the one vertical collection is using to render. Part of the design of those events (afaik) is that they should *not* be fired synchronously. Instead they should fire on an idle frame (aka `setTimeout` for older browsers).

Thus I think the appropriate fix here might better be:

a) close over the `item` or key at schedule-time instead of at event fire-time
b) if firing an event where the item is no longer in `items`, cancel the firing

Emerson:
@andreyfel - is there a path forward on this PR? Would be really nice to get this fix landed in a formal VC release.

runspired:
I think @mixonic's take is correct, the timing of the run loop we do here is *very* intentional for perf. We should likely do option "b" though both options seem like good ideas to do are complementary.

andreyfel:
We switched to a different fix, see #359.

PR #326, opened 2021-02-08
Drop 2.4

twokul:


.github/workflows/ci.yml
```diff
@@ -42,9 +42,6 @@ jobs:
         ember-version:
           [
             # The test suite only support Ember 3.x
-            # ember-lts-2.8,
-            # ember-lts-2.12,
-            # ember-lts-2.18,
```
mixonic, line 47: Can we leave this? I'd like to express the ambition that the suite works with them.

Having poked at dependency issues for a few days now, I think I also have an idea as to what could be breaking and it may just be in the test suite.

.travis.yml
```diff
@@ -69,7 +67,7 @@ before_install:
   - yarn global add greenkeeper-lockfile@1
 
 install:
-  - yarn install --non-interactive
+  - yarn install --non-interactive --frozen-lockfile
```
mixonic, line 70: I believe for addons you want to *not* frozen-lockfile. The idea is that when a consuming app installs your dependencies, they get the latest matching versions not the contents of your lockfile. Thus to really test for those consuming cases, you need let the dependencies float.

It does mean your devDeps float in addition to your dependencies, however.

mixonic (changes requested):

mixonic (approved):

Issue #320, opened 2020-05-14
Application test times out when two vertical collections share a containerSelector


steveszc:
Minimal reproduction repo: https://github.com/steveszc/vertical-collection-shared-container-reproduction

When a route template contains two vertical-collections that share a common containerSelector (ex a scrollable `body`) a test that visits that route will timeout on the `await visit()`. The vertical-collections work as expected when using the app, however.

steveszc:
Maybe related: https://github.com/html-next/ember-raf-scheduler/issues/5

steveszc:
Related: https://github.com/html-next/vertical-collection/issues/197

steveszc:
I spent some time debugging with some teammates. We were able to identify the root cause of this issue, which is a faulty conditional inside the scroll handler util. 
TL;DR; If multiple scroll handlers are registered to the same element AND passive scroll listeners are supported, then any time after the first scroll handler is added the addScrollHandler method would begin polling (even though it already added a passive scroll listener). This extra polling would trigger an extra job to be scheduled with ember-raf-scheduler, but this extra job was unexpected/accidental and was never flushed, which caused the ember-raf-scheduler test waiter to get stuck and never resolve.  
https://github.com/html-next/vertical-collection/blob/master/addon/-private/data-view/utils/scroll-handler.js#L53 

I'll plan to submit a PR with the fix and a test in the next week or two.

MariannaAtPlay:
Hello!
Are there any plans to fix this issue? We are running into this as well!
Thank you @steveszc for figuring out the root cause.

PR #327, opened 2021-02-08
avoid concurrent passiveHandler and poll

ro0gr:
...for a single scroll container

fixes https://github.com/html-next/vertical-collection/issues/320

like it's stated, around the relevant code, it's missing code coverage, I'd try figure out a way to test it.

todo:

 - [ ] tests

### DIFF
--- a/addon/-private/data-view/utils/scroll-handler.js
+++ b/addon/-private/data-view/utils/scroll-handler.js
@@ -36,26 +36,22 @@ export class ScrollHandler {
         handlers
       };
       // TODO add explicit test
-      if (SUPPORTS_PASSIVE) {
+      if (this.isUsingPassive) {
         cache.passiveHandler = function() {
           ScrollHandler.triggerElementHandlers(element, cache);
         };
+
+        element.addEventListener('scroll', cache.passiveHandler, { capture: true, passive: true });
+      // TODO add explicit test
       } else {
         cache.passiveHandler = UNDEFINED_VALUE;
+
+        this.poll();
       }
     } else {
       cache = this.handlers[index];
       handlers = cache.handlers;
       handlers.push(handler);
-    }
-
-    // TODO add explicit test
-    if (this.isUsingPassive && handlers.length === 1) {
-      element.addEventListener('scroll', cache.passiveHandler, { capture: true, passive: true });
-
-    // TODO add explicit test
-    } else if (!this.isPolling) {
-      this.poll();
     }
   }
 


addon/-private/data-view/utils/scroll-handler.js
```diff
@@ -36,27 +36,23 @@ export class ScrollHandler {
...
+        this.poll();
       }
     } else {
       cache = this.handlers[index];
       handlers = cache.handlers;
       handlers.push(handler);
     }
-
-    // TODO add explicit test
-    if (this.isUsingPassive && handlers.length === 1) {
```
ro0gr, line 53: this block was moved to a similar `if/else` above, just in order to cleanup a bit.

However, since I did it some time ago, I don't remember exactly why I've ignored the `handlers.length === 1` part. Will need to take another look at it.

addon/-private/data-view/utils/scroll-handler.js
```diff
@@ -36,27 +36,23 @@ export class ScrollHandler {
         handlers
       };
       // TODO add explicit test
-      if (SUPPORTS_PASSIVE) {
+      if (this.isUsingPassive) {
```
ro0gr, line 39: in the constructor there is

```js
this.isUsingPassive = SUPPORTS_PASSIVE;
```

I guess, this way there is a chance to configure  `isUsingPassive` in tests.

addon/-private/data-view/utils/scroll-handler.js
ro0gr: So the source of error was exactly the `handlers.length === 1` check.

When we add the second handler to the same scrollable target, we fallen back to the polling mode, which in its turns makes test waiters to freeze.

tests/unit/-private/data-view/utils/scroll-handler-test.js
```diff
@@ -192,11 +243,68 @@ test('Multiple elements with handlers works as expected', (assert) => {
...
+  assert.strictEqual(scrollable1Index, 0, `The scrollable was added to the watched elements list.`);
+
+  let cache1 = scrollHandlers.handlers[0];
+
+  assert.true(cache1.handlers.length === 2, `We added the handler`);
+
+  // test triggering that handler
+  assert.strictEqual(scrollable.scrollTop, 0, `The scrollable is initially unscrolled`);
+
+  assert.false(scrollHandlers.isPolling, 'polling is inactive');
```
ro0gr, line 285: this is the only assert which is fixed by this PR

PR #330, opened 2021-02-09
Remove targets file

twokul:
One thing to note is that the default configuration is a little different from what we had before. [Link to targets doc](https://cli.emberjs.com/release/advanced-use/build-targets/). It will not impact the consumers of the library since "dummy" application is used for testing purposes.

```javascript
// config/targets.js
module.exports = {
  browsers: [
    'last 1 edge versions',
    'last 1 Chrome versions',
    'last 1 Firefox versions',
    'last 1 Safari versions'
  ]
};
```


rwwagner90 (changes requested):
This seems to have broken tests. Could we just set it to the default blueprint values, since it's only the dummy app anyway?

PR #331, opened 2021-02-10
Restore support for Ember 2.x

mixonic:
`master` of the vertical-collection repo had some Ember-pre-3.x
compatibility stuff removed back in 2019. For example:

* https://github.com/html-next/vertical-collection/commit/011d9a48272e80cf8e76df9be383244615597a09

Restore one of those changes.

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -18,8 +18,15 @@ module.exports = async function() {
         },
         npm: {
           devDependencies: {
-            'ember-source': null
+            'ember-source': null,
+            'ember-factory-for-polyfill': '1.3.1'
           },
+          dependencies: {
+            'ember-compatibility-helpers': '1.2.1'
+          },
+          resolutions: {
+            'ember-compatibility-helpers': '1.2.1'
+          }
         },
       },
       {
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -33,6 +33,7 @@ module.exports = async function() {
         name: 'ember-lts-2.12',
         npm: {
           devDependencies: {
+            '@ember/jquery': '^1.1.0',
             'ember-source': '~2.12.0'
           },
         },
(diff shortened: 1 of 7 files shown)


mixonic:
As @rwwagner90 reviewed https://github.com/html-next/vertical-collection/pull/321 he noted that the test suite was not running for Ember 2.x. I replied that it seemed like the 2.x test suite was just broken, indeed master generally is in a poor state.

Despite there being some 1.13 compatibility code we found and dropped as part of https://github.com/html-next/vertical-collection/pull/321, there was also an earlier effort to drop some compatibility code from `master`, for example in https://github.com/html-next/vertical-collection/pull/289 (as pointed out by @rwwagner90 over DM). Based on us still finding compat code on master, I had presumed that master still should support 2.x, but I was wrong!

In this PR I've restored the Ember 2.x compatibility code and massaged the ember try config until we have a suite passing through all of 2.x.

config/ember-try.js
```diff
@@ -18,14 +18,22 @@ module.exports = async function() {
         },
         npm: {
           devDependencies: {
-            'ember-source': null
+            'ember-source': null,
+            'ember-factory-for-polyfill': '1.3.1'
           },
+          dependencies: {
+            'ember-compatibility-helpers': '1.2.1'
```
rwwagner90, line 25: Do we need this in ember-try since it's a dependency of the addon already now?

rwwagner90 (approved):

config/ember-try.js
mixonic: Yes, this is required. I confirmed it by dropping it locally and running the try config.

The dependency in the `package.json` is `^1.2.1`, where this is `1.2.1`. Wait wait, I agree, the `resolutions` should take care of that, however practically it doesn't do the trick on its own.

PR #335, opened 2021-09-20
ember-cli 3.28, ember-cli-htmlbars 5.x, require Ember >= 3.8

rwwagner90:
The ember-cli-htmlbars update to 5.x is incompatible with Ember < 3.8, so this is a breaking change in Ember version support.

PR #343, opened 2021-12-07
Update CI to Ember 2.18+

mixonic:
A patch to start v3 on `master`. Will support Ember 2.18.0 through latest, will support Node 14+.

If this lands, we should also start a `release-2` branch for patches back to 2.x.


mixonic:
Cool, this is basically proven out. With no real changes other than bumping `ember-cli-htmlbars`, Ember 3.28 support (for CI at least) is unblocked. I'd like to look through https://github.com/html-next/vertical-collection/pull/335 and see what can be backported to see if we can retain Ember 2.18 support while unblocking Ember 4. It seems possible.

mixonic:
Alright! I've gone through a number of failures and looked at https://github.com/html-next/vertical-collection/pull/335 in detail as well. This is getting close though I've still got a failure.

This evening I backported four patches which can safely land on `master` (and even make `master` / v2 a bit nicer for anyone using them in later Ember versions). Those patches are:

* https://github.com/html-next/vertical-collection/pull/345
* https://github.com/html-next/vertical-collection/pull/346
* https://github.com/html-next/vertical-collection/pull/347
* https://github.com/html-next/vertical-collection/pull/348
* Earlier I landed https://github.com/html-next/vertical-collection/pull/344 which also make a change to an import path (`htmlSafe`) to be future-safe.

The content of what remains in this branch/patch is:

* Adjust CI, `package.json` and `ember-try` etc to drop < 2.18 and add 4.0+ / latest
* Port @chancancode 's patch https://github.com/html-next/vertical-collection/pull/337 to fix the internal rollup build
* Re-add a part of https://github.com/html-next/vertical-collection/pull/334 which I omitted in https://github.com/html-next/vertical-collection/pull/345 (use of `@arg`)
* Drop `ember-template-lint`. I do not believe it is compatible with Ember 4 (it contains a transform/lint which doesn't return a node, and that behavior is unsupported in Ember 4), but I haven't tried to reproduce that upstream.
* Pin https://github.com/html-next/ember-raf-scheduler/pull/12. That needs to go green upstream and be released, but it is basically only dropping a global `Ember` use.
* Upgrade the test suite :grimacing: This is the ugly bit, however I think it is required for Ember 4 compatibility (I can re-assert this easily and will do so, I think the the issue is some old helpers and such use global Ember). I think https://github.com/html-next/vertical-collection/pull/335 must has something I'm missing here as it is green for the 3.x builds and this is not. Clearly this is a blocker.

mixonic:
Ok, green and landing it. I'll cut `3.0.0-0` from `master` after the merge.

Thanks @rwwagner90 @chancancode @johanrd for the illustrative work on https://github.com/html-next/vertical-collection/pull/335. That was a powerful guide. I think most of the critical components from that PR are represented here, the biggest difference in lines being that this PR doesn't attempt to change the linting and code style. I'd love to work with you on getting any of that landed as incremental changes on the main branch.

mixonic:
`3.0.0-0` has been released, and a `release-2` branch for legacy v2 support has been created.

PR #350, opened 2021-12-09
Use modern link-to API

mixonic:


mixonic:
This approach works for ember 2.8+, so moving it into https://github.com/html-next/vertical-collection/pull/343

Issue #357, opened 2022-05-23
Bug: Updating deleted items


Atrue:
After removing the item from the list the item's node can still exist in the virtual `domPool` element so its component isn't destroyed and can be updated by Ember.

This behavior isn't working for computed fields on the ember models. Ember rejects the error when trying to access the computed property on the destroyed object if it hasn't been consumed before. 
```
Assertion Failed: Attempted to access the computed `computedProperty` on a destroyed object, which is not allowed.
```
Expected behavior: if the item is removed from the items list, this node shouldn't be updated or should be removed.

PR #358, opened 2022-05-23
Delete virtual element

Atrue:
Issue - https://github.com/html-next/vertical-collection/issues/357
See [test updating deleted elements](https://github.com/html-next/vertical-collection/pull/358/commits/39571e5d519c963d161819dd92d19a3bf964a273) commit to reproduce the issue in tests

**Description**:
The radar has a domPool virtual element to save temporarily deleted nodes and reuse them when a new item is added. So if the item is deleted from the items list the item's node can still exist in the virtual dom so the item's component can be still updated by Ember.

This behavior isn't working for computed fields on the ember models. Ember rejects the error when trying to access the computed property on the destroyed object if it hasn't been consumed before. This is possible if the items' component should render more fields after the update.

To avoid re-rendering deleted items it's better to delete related components from the virtual pool. It doesn't give much impact on performance as this can happen only if the component has fewer total items than virtual items after the update and it also doesn't affect scroll rendering.

### DIFF
--- a/addon/-private/data-view/radar/radar.js
+++ b/addon/-private/data-view/radar/radar.js
@@ -1,6 +1,7 @@
 import { A } from '@ember/array';
 import { set, get } from '@ember/object';
 import { assert } from '@ember/debug';
+import { run } from '@ember/runloop';
 import { DEBUG } from '@glimmer/env';
 
 import { Token, scheduler } from 'ember-raf-scheduler';
@@ -530,11 +531,19 @@ export default class Radar {
     if (_componentPool.length > 0) {
       if (shouldRecycle === true) {
         // Grab the DOM of the remaining components and move it to temporary node disconnected from
-        // the body. If we end up using these components again, we'll grab their DOM and put it back
-        for (let i = 0; i < _componentPool.length; i++) {
+        // the body if the item can be reused later otherwise delete the component to avoid virtual re-rendering of the
+        // deleted item. If we end up using these components again, we'll grab their DOM and put it back
+        for (let i = _componentPool.length - 1; i >= 0; i--) {
           const component = _componentPool[i];
-
-          insertRangeBefore(this._domPool, null, component.realUpperBound, component.realLowerBound);
+          const item = objectAt(items, component.index);
+          if (item) {
+            insertRangeBefore(this._domPool, null, component.realUpperBound, component.realLowerBound);
+          } else {
+            run(() => {
+              virtualComponents.removeObject(component);
+            });
+            _componentPool.splice(i, 1);
+          }
         }
       } else {
         virtualComponents.removeObjects(_componentPool);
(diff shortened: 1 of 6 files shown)


addon/-private/data-view/radar/radar.js
```diff
@@ -530,11 +531,19 @@ export default class Radar {
...
+        // deleted item. If we end up using these components again, we'll grab their DOM and put it back
+        for (let i = _componentPool.length - 1; i >= 0; i--) {
           const component = _componentPool[i];
-
-          insertRangeBefore(this._domPool, null, component.realUpperBound, component.realLowerBound);
+          const item = objectAt(items, component.index);
+          if (item) {
+            insertRangeBefore(this._domPool, null, component.realUpperBound, component.realLowerBound);
+          } else {
+            run(() => {
```
Atrue, line 542: Deleting all virtual elements in the same runloop gives unexpected results in Ember >= 3.20. Please write if you need some extra details

tests/acceptance/acceptance-tests/record-array-test.js
```diff
@@ -13,4 +13,16 @@ module('Acceptance | Record Array', function(hooks) {
...
+
+  test('RecordArrays updates correctly after deleting items', async function(assert) {
+    await newVisit('/acceptance-tests/record-array');
+
+    assert.equal(findAll('number-slide').length, 15, 'correct number of items rendered');
+    await click('#update-items-button');
+    assert.equal(findAll('number-slide').length, 5, 'correct number of items rendered');
+    assert.deepEqual(findAll('number-slide').map(s => s.textContent.trim()[0]), ['0', '1', '2', '3', '4'], 'correct items order');
+    await click('#show-prefixed-button');
+    assert.equal(findAll('number-slide').length, 5, 'correct number of items rendered and nothing crushes');
```
mixonic, line 25:
```suggestion
    assert.equal(findAll('number-slide').length, 5, 'correct number of items rendered and nothing crashes');
```

?

addon/-private/data-view/radar/radar.js
mixonic: `removeObject` should be in a runloop :+1:

addon/-private/data-view/radar/radar.js
```diff
@@ -530,11 +531,19 @@ export default class Radar {
...
+              virtualComponents.removeObject(component);
+            });
+            _componentPool.splice(i, 1);
```
mixonic, line 545: Ah, the reverse traversal is so this mutation is safe, yes?

The `insertRangeBefore(` logic doesn't care about the order of traversal? I take it the answer is no, but I'm not so familiar with the code that I want to presume it without diving in.

mixonic (commented):
This looks very reasonable, and like it was tricky to resolve. Bravo.

I left just a few queries and kicked off the test suite.

addon/-private/data-view/radar/radar.js
Atrue: yep, as the indexes are only decreasing you are sure you don't miss any item of the _componentPool in the loop.
There is no sense in order in `insertRangeBefore` as the `_componentPool` is shuffled anyway

mixonic:
@atrusov-retailnext thanks for this work. I'll get it into a minor release shortly.

runspired:
released in 3.1

PR #359, opened 2022-05-26
Don't send action on non-existing items

azhiv:
This is a straightforward fix for https://github.com/html-next/vertical-collection/issues/317.
We tried a more sophisticated fix https://github.com/html-next/vertical-collection/pull/324, but the error didn't vanish - it rather started to appear more seldom. So eventually we decided to perform this check on the `item` which utterly fixes the issue.
- Check that the item the event is being generated for still exists.


andreyfel:
@runspired @mixonic can you guys take a look?

mixonic:
Addressed by https://github.com/html-next/vertical-collection/pull/391

Thank you @azhiv!

Issue #362, opened 2022-06-03
v4 Proposal


mixonic:
Following https://github.com/html-next/vertical-collection/issues/341, I propose a 4.0 branch and release.

Goals:

* Drop support for Ember < 3.12-LTS.
* v2 Ember addon? Is this even possible with the above constraint? I'm not sure. Maybe a nice to have. See https://github.com/html-next/vertical-collection/issues/260
* No change in Node support.
* Release 1.0 of https://github.com/html-next/ember-raf-scheduler.

Issue #260, opened 2019-05-15
refactor treeForAddon for embroider compatibility


efx:
As a user of vertical-collection, I would like to avoid using a workaround when building my application with [embroider](https://github.com/embroider-build/embroider).

See @ef4's helpful analysis and workaround here:
https://github.com/embroider-build/embroider/issues/169#issuecomment-491953938

mixonic:
I suggest this be considered as part of a 4.0 effort https://github.com/html-next/vertical-collection/issues/362

PR #372, opened 2022-07-20
Vertical Collection 4

mixonic:
Satisfying the proposal in https://github.com/html-next/vertical-collection/issues/362, start implementing a v4 which:

* Drops support for Ember < 3.12-LTS.
* Drops support for Ember CLI 2.x.
* v2 Ember addon? Is this even possible with the above constraint? I'm not sure. Maybe a nice to have. See
https://github.com/html-next/vertical-collection/issues/260
* No change in Node support.
* ~Release 1.0 of [html-next/ember-raf-scheduler](https://github.com/html-next/ember-raf-scheduler).~ Ember 4.5 support is confirmed in https://github.com/html-next/vertical-collection/pull/373, so I don't think we actually need to do this.
* Drop the positional param for `items` on the vertical collection component.

Please open PRs against `master` if they are backward compatible to Vertical Collection 3, or against this branch if they must target v4.


mixonic:
`v4.0.0-beta.0` has been released to NPM https://www.npmjs.com/package/@html-next/vertical-collection/v/4.0.0-beta.0

PR #377, opened 2022-07-21
Add a QUnit level test timeout

mixonic:


mixonic:
Although this seems fine, it doesn't prevent the test runs from hanging.

PR #380, opened 2022-08-04
fix: enable parallel builds

runspired:
tests will also catch regressions now that the flag is set to be on for our own builds. This is a non-breaking change and can be released as a patch.


runspired:
The builds are hanging in CI but passing locally. I'm going to open an empty PR I suspect master is simply in a bad state.

Issue #382, opened 2022-08-04
Tests are flakey and unstable


runspired:
tests will fail in unpredictable patterns, and this is somewhat hidden by CI timing out. Obervable in #381 where a readme change leads to failures.

mixonic:
There are two issues:

* First, there are unpredictable failures in the test suite. These are papered over by a retry script which re-runs the whole suite on CI if there is a failure, and only gives up after several attempts.
* Second, there is something in the test suite or runner which causes the github action to hang. This second item is what the timeout circumvents. Without a timeout, it simply hangs for hours.

PR #384, opened 2022-08-04
Lower a CI timeout back to 9 minutes

mixonic:
There is something in the vertical collection test suite which periodically hangs the action runner. Extending the timeout to 15 minutes doesn't make it more likely we get a green build, instead it increases the delay you need to wait before retrying.

Relates to #382


runspired:
After playing around locally we might actually consider an even shorter timeout. What I think led me astray is that increasing it caused the failures to actually report vs just the hangs.

PR #388, opened 2022-09-01
[v4] Refactor a bunch of debug code to DEBUG

mixonic:
Previously this code used a custom compilation step to be included. Move it over to the standard DEBUG flags already stripped by other tooling.


mixonic:
@twokul it is green 😬 Seems good to land?

PR #391, opened 2022-10-25
clear scheduled actions after recalculation

Atrue:
Fixes https://github.com/html-next/vertical-collection/issues/317. Similar https://github.com/html-next/vertical-collection/pull/359 but using another approach + tests

Scheduled actions as the name says are fired with delay after the render is finished. They are going through the `sync` queue (`scheduleUpdate` method) and the `measure` queue (`update` method) in the radar class and `setTimeout` (`_scheduleSendAction` method) in the component. So there is some gap after the render is finished and before the event is actually fired when the `items` argument can be changed and the event will be fired with an exception. The scheduled actions include the event and index of the item, but due to timeout, the actual item may not exist for this index. 

Clear the timeout and the queue before scheduling a new update after recalculation to omit these outdated events and fire only the new actual events

### DIFF
--- a/addon/components/vertical-collection/component.js
+++ b/addon/components/vertical-collection/component.js
@@ -327,12 +327,19 @@ const VerticalCollection = Component.extend({
     _radar.bufferSize = this.bufferSize;
 
     _radar.scheduleUpdate(true);
+    this._clearScheduledActions();
 
     return _radar.virtualComponents;
   }),
 
   schedule(queueName, job) {
     return scheduler.schedule(queueName, job, this.token);
+  },
+
+  _clearScheduledActions() {
+    clearTimeout(this._nextSendActions);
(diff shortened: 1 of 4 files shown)


mixonic (approved):
This looks goo to me! Thank you for taking the time to isolate the issue down to a test.

I'm babysitting the build to get it green.

andreyfel:
@mixonic will you be able to cut a release with this fix? Thanks!

mixonic:
Of course. I'm running it through some full-stack checks. See https://github.com/Addepar/ember-table/pull/969. I'm also running it on our app at Addepar as a sanity check :+1:

mixonic:
Now released in 4.0.1 :tada: Thank you @Atrue and team.

andreyfel:
Thanks @mixonic!

PR #396, opened 2022-11-11
Fix build for 4.0

### LINKED ISSUE #393: Transpiling error in Ember 4 without embroider
The below only happens when not using embroider.

Line 352 of radar.js
```
    assert('Must provide a `estimateHeight` value to vertical-collection', estimateHeight !== null);
```
is being transpiled to
```
      (!(estimateHeight !== null) && Ember.assert('Must provide a `estimateHeight` value to vertical-collection', estimateHeight !== null));
```
This causes the error `Ember is not defined` 

[package.json](https://github.com/html-next/vertical-collection/files/9952854/package.json.txt)

wagenet:
Fixes #393.

...or at least attempts to!

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,17 +19,6 @@ module.exports = {
     return '';
   },
 
-  // Borrowed from ember-cli-babel
-  _emberVersionRequiresModulesAPIPolyfill() {
-    let checker = this.checker.for('ember-source', 'npm');
-
-    if (!checker.exists()) {
-      return true;
-    }
-
-    return checker.lt('3.27.0-alpha.1');
-  },
-
   treeForAddon(tree) {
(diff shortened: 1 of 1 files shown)


mixonic:
FYI CI on the repo is very bad. I may need to rerun failures to get then green, if they don't look related.

runspired:
Best I can tell CI failures are unrelated for most of the versions but we may have broken 3.12/3.16/3.20

mixonic:
Uh. I hope we did not? Those versions are supported and the support is used by ember table and apps I work on.

runspired:
@mixonic you missed I added a commit to fix them

mixonic:
Ah, amazing. Thanks @runspired @wagenet !

PR #401, opened 2023-08-25
Drop 3.12 and earlier from CI

mixonic:


mixonic:
Replaced by https://github.com/html-next/vertical-collection/pull/402

PR #402, opened 2023-09-08
v3.28.0...v5.2.0

mixonic:
* Upgrade Ember CLI from 3.28 -> 5.2
* Upgrade Ember source from 3.12 -> 5.2
* Upgrade Ember data from 3.12 -> 4.12

Also

* Adds Volta
* Bumps Node to v16
* Drop Ember versions < 3.28 from the CI
* Add Ember 5 to the CI
* Address many code style issues after the Ember CLI upgrade


CONTRIBUTING.md
```diff
@@ -3,7 +3,7 @@
 ## Installation
 
 * `git clone <repository-url>`
```
runspired, line 5:
```suggestion
* `git clone <repository-url> html-next-vertical-collection`
```

runspired: without this the `cd` command below doesn't make sense

index.js
```diff
@@ -62,8 +62,6 @@ module.exports = {
         disableDecoratorTransforms: false,
         enableTypeScriptTransform: true,
 
-        throwUnlessParallelizable: true,
```
runspired, line 65: any reason we didn't leave this? if we aren't parallelizable something bad has gone wrong. I've worked pretty hard to keep things parallelizable around the ecosystem opening fixes when hitting issues.

tests/dummy/app/index.html
```diff
@@ -2,7 +2,6 @@
 <html>
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
runspired, line 5: thank god we've made it past this era

tests/dummy/app/routes/application/template.hbs
```diff
@@ -1,8 +1,8 @@
 <header class="container-fluid">
   <div class="row">
     <h1 class="col-sm-6">
-      <LinkTo @route="index">
-        <img alt="HTML Next Logo" src="./HTML-Next.png" style="height: 2em;">
+      <LinkTo @route="index" style="color:#222">
```
runspired, line 4: why do we end up needing the extra color?

tests/dummy/app/routes/examples/infinite-scroll/controller.js
```diff
@@ -10,21 +10,21 @@ export default Controller.extend({
...
+      let model = this.model.data.numbers;
       model.unshiftObjects(numbers);
       // this.set('model.numbers', newModel);
-      this.set('model.first', first - 20);
+      this.set('model.data.first', first - 20);
     },
 
     loadBelow() {
-      let last = this.model.last;
+      let last = this.model.data.last;
```
runspired, line 22: I am sorry for the sins that I committed

tests/integration/modern-ember-test.js
```diff
@@ -23,8 +22,6 @@ module('vertical-collection', 'Integration | Modern Ember Features Tests', funct
         </div>
       `);
 
-    await settled();
```
runspired, line 26: note: I dislike this lint rule as its actually dishonest. But also yay cleanup

runspired (approved):
no blocking feedback, left a few nits inline.

tests/dummy/app/routes/application/template.hbs
mixonic: The link is an `a` tag and turns blue, but this is the logo on the top left in the header. We want to to be the same color as the logo image, which is `#222`.

This was probably broken in https://github.com/html-next/vertical-collection/commit/3343ecc3506c23c6ba10ed18a3f915672f5ff8c7#diff-a6b32449eac6431ef7263de086a744f8aed3883a686fe93880e607f5e08ffdb9L3, although I don't believe the docs site after that change was ever deployed (it doesn't seem like it, at least).

I agree this is a bit lame, but I think you agree we have bigger fish to fry.

index.js
mixonic: I wasn't able to trace the error back to the root cause. I can give it another look, but I'm going to consider this non-blocking and decouple the investigation. Frankly I considered this option as being old enough that I wasn't sure it was still something meaningful.

A conversion of this whole project to a v2 addon would simplify the dependencies and I presume make the build parallelization concerns moot (at least for consumers of this library), and I suspect that is the meta.

PR #403, opened 2023-09-11
Rewrite VerticalCollection to a GlimmerComponent

mixonic:


addon/components/vertical-collection/component.js
```diff
@@ -308,48 +318,68 @@ const VerticalCollection = Component.extend({
...
+  }
 
-  virtualComponents: computed('items.[]', 'renderAll', 'estimateHeight', 'bufferSize', function() {
-    const { _radar } = this;
+  get shouldYieldToInverse() {
+    return this.isEmpty;
+  }
 
-    const items = this.items;
+  @cached
```
mixonic, line 351: I'm not convinced this observes the contents of `items.[]` as it did previously. The most compatible fix may simply be to return it to being a CP

addon/components/vertical-collection/template.hbs
```diff
@@ -1,3 +1,4 @@
+{{vertical-collection/call this.templateDidRender}}
```
mixonic, line 1: This is basically a hack to get `didInsertElement` timings for a callback.

addon/components/vertical-collection/component.js
mixonic: In the current iteration of this PR, I've returned this to a computed

PR #408, opened 2025-01-10
Bump dev Node to v18, reroll lockfile

mixonic:
`master` CI fails due to some package now requiring Node 18. Resolve this by bumping the dev-pinned Node version to 18. Additionally reroll the lockfile.


mixonic:
Ember 6.x still fails the build, but I believe this is due to something in the test app. Landing this bump as incremental progress.

PR #438, opened 2025-05-08
Move to pnpm

NullVoxPopuli:
yarn is silly


package.json
```diff
@@ -112,11 +112,11 @@
     "ember-source": ">= 3.28.0"
   },
   "engines": {
-    "node": "16.* || >= 18"
+    "node": ">= 18"
```
NullVoxPopuli, line 115: this _would_ be a breaking change if we cared about node support -- but my hope is to have vertical collection be a v2 addon before the next release so node version just won't be relevant at that point

PR #522, opened 2026-01-26
Update raf-scheduler

NullVoxPopuli:
Trying out inlining https://github.com/html-next/ember-raf-scheduler/blob/v0.5.0-ember-raf-scheduler/src/index.js
To see if any issues need to be fixed.

If not, we'll need another breaking change probably to set the min-version to 0.5.0.
Otherwise, with inlining, we may not need a breaking change

See also:
- https://github.com/html-next/ember-raf-scheduler/pull/22


johanrd:
maybe stale? raf-scheduler seems to be at 0.5.1 now:
https://github.com/html-next/vertical-collection/blob/f01c770f1dde915394651eb2a0635df6ae77e9cc/vertical-collection/package.json#L65

PR #561, opened 2026-04-09
Upgrade raf scheduler to 0.5.1

NullVoxPopuli:


NullVoxPopuli:
Merging without green CI, because of upstream issue with warp-drive's compatibility: https://github.com/warp-drive-data/warp-drive/issues/10510

NullVoxPopuli-ai-agent:
The lockfile has a stale resolution for the `vertical-collection` workspace package — it resolves `ember-raf-scheduler` to **0.3.0** instead of 0.5.1:

```yaml
'@html-next/vertical-collection@file:vertical-collection':
    dependencies:
      '@embroider/addon-shim': 1.10.2
      ember-raf-scheduler: 0.3.0    # <-- should be 0.5.1
```

0.3.0 doesn't have the `_flush` waiter token fix, so `settled()` returns before pending RAF work completes → PromiseArray tests timeout.

The ember-try scenarios and "Floating Dependencies" pass because they re-resolve dependencies (getting 0.5.1). Only the base "Tests" job uses the stale lockfile.

**Fix:** re-roll the lockfile (`rm -rf node_modules vertical-collection/node_modules test-app/node_modules pnpm-lock.yaml && pnpm install`).

PR #569, opened 2026-09-22
chore: bump warp-drive and embroider to fix ember 7 CI

johanrd:
`ember-release`, `ember-beta` and `ember-canary` have been failing on `main` since ember-source reached 7.x — see [run 33611545603](https://github.com/html-next/vertical-collection/actions/runs/33611545603). They fail during ember-try setup, before any test runs:

```
[vite]: Rollup failed to resolve import "ember" from
node_modules/.../@warp-drive/legacy/dist/compat/extensions.js
```

`@warp-drive/legacy@5.8.1` has `import Ember from 'ember'` on line 3 of that file, and Ember 7 no longer provides the `ember` barrel module. 5.9.1 has no bare `ember` import left anywhere in its dist.

`@warp-drive/*` and `@ember-data/*` were already declared as `^5.8.1`, so only the lockfile held them back; the ranges move to `^5.9.1` to record the minimum that works on Ember 7. `eslint-plugin-warp-drive` was pinned exactly and moves with them.

The embroider bump comes along for one reason: warp-drive 5.9 emits `macroCondition(...) && ...`, which `@embroider/macros` only accepts from 1.21 onward ("*macroCondition can only be used as the predicate of an if statement, ternary expression, or && / || logical expression*"). Bumping `@embroider/macros` alone is not enough, because `@embroider/compat`/`core`/`vite` carry their own transitive copy that runs the babel plugin.

Verified locally:

- `ember try:one ember-release` → **SUCCESS** on ember-source `7.3.0-release+4bffdcad`
- `pnpm --filter test-app test` → 348/348 pass on the default scenario
- `pnpm lint` clean, including the new `eslint-plugin-warp-drive` version

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mixonic (approved):
Thank you!

